PR #5, opened 2020-01-07
IP whitelist

thetonymaster:
Support for Project IP whitelist for CFN


examples/project-ip-whitelist/project-ip.json
```diff
@@ -0,0 +1,37 @@
+{
+  "AWSTemplateFormatVersion" : "2010-09-09",
+  "Description" : "This template creates a Project and a Database User on the MongoDB Atlas API, this will be billed to your Atlas account.",
```
shum: Description looks incorrect

ip-whitelist/cmd/resource/resource.go
```diff
@@ -0,0 +1,247 @@
...
+
+// Create handles the Create event from the Cloudformation service.
+func Create(req handler.Request, prevModel *Model, currentModel *Model) (handler.ProgressEvent, error) {
+	client, err := util.CreateMongoDBClient(*currentModel.ApiKeys.PublicKey.Value(), *currentModel.ApiKeys.PrivateKey.Value())
+	if err != nil {
+		return handler.ProgressEvent{}, err
+	}
+
+	projectID := *currentModel.ProjectId.Value()
+
```
shum: [nit] remove new line

ip-whitelist/cmd/resource/resource.go
```diff
@@ -0,0 +1,247 @@
...
+	request := getProjectIPWhitelistRequest(currentModel)
+
+	_, _, err = client.ProjectIPWhitelist.Create(context.Background(), projectID, request)
+	if err != nil {
+		return handler.ProgressEvent{}, err
+	}
+
+	var withelist []string
```
shum: [typo] `whitelist`

ip-whitelist/cmd/resource/resource.go
```diff
@@ -0,0 +1,247 @@
...
+	return whitelist
+}
+
+func whiteListMap(whitelist []*mongodbatlas.ProjectIPWhitelist, f func(string)) {
+	for _, entry := range whitelist {
+		if entry.CIDRBlock != "" {
+			f(entry.CIDRBlock)
+		} else if entry.IPAddress != "" {
+			f(entry.IPAddress)
+		}
```
shum: @themantissa I think we added support for AWS security groups in https://github.com/mongodb/go-client-mongodb-atlas/pull/35 do we want to do the same here?

ip-whitelist/cmd/resource/resource.go
```diff
@@ -0,0 +1,247 @@
...
+		if w.CidrBlock != nil {
+			wl.CIDRBlock = *w.CidrBlock.Value()
+		}
+		if w.IpAddress != nil {
+			wl.IPAddress = *w.IpAddress.Value()
+		}
+
+		whitelist = append(whitelist, wl)
+
```
shum: [nit] remove new line

ip-whitelist/cmd/resource/resource.go
```diff
@@ -0,0 +1,247 @@
...
+	if err != nil {
+		return handler.ProgressEvent{}, err
+	}
+
+	var withelist []string
+	whiteListMap(request, func(entry string) {
+		withelist = append(withelist, entry)
+	})
+
+	id := encodeStateID(map[string]string{
```
shum: Could you explain to me the technical rationale of why we encode the id and the entries into a giant string? It's a weird approach so just trying to understand why we do it.

ip-whitelist/cmd/resource/resource.go
```diff
@@ -0,0 +1,247 @@
...
+}
+
+func encodeStateID(values map[string]string) string {
+	encode := func(e string) string { return base64.StdEncoding.EncodeToString([]byte(e)) }
+	encodedValues := make([]string, 0)
+
+	for key, value := range values {
+		encodedValues = append(encodedValues, fmt.Sprintf("%s:%s", encode(key), encode(value)))
+	}
+	return strings.Join(encodedValues, "-")
```
shum: If we do add support for AWS SG as above, I think we'll have to change this delimiter

ip-whitelist/cmd/resource/resource.go
```diff
@@ -0,0 +1,247 @@
...
+			wl.IPAddress = *w.IpAddress.Value()
+		}
+
+		whitelist = append(whitelist, wl)
+
+	}
+	return whitelist
+}
+
+func whiteListMap(whitelist []*mongodbatlas.ProjectIPWhitelist, f func(string)) {
```
shum: weird camel case, suggest lowercasing the `L` to `whitelistMap`

ip-whitelist/cmd/resource/resource.go
```diff
@@ -0,0 +1,247 @@
...
+
+	ids := decodeStateID(*currentModel.Id.Value())
+
+	whitelist, err := getProjectIPWhitelist(ids, client)
+	if err != nil {
+		return handler.ProgressEvent{}, err
+	}
+
+	whiteListMap(whitelist, func(entry string) {
+		_, err = client.ProjectIPWhitelist.Delete(context.Background(), ids["project_id"], entry)
```
shum: Maybe I missed the larger discussion of this that happened in another repo. Why do we opt to manually delete + recreate the whitelist entry, rather than rely on the update behavior of https://docs.atlas.mongodb.com/reference/api/whitelist-update-one/ ?

ip-whitelist/mongodb-atlas-projectipwhitelist.json
```diff
@@ -0,0 +1,106 @@
...
+                "ipAddress": {
+                    "description": "Whitelisted IP address. Mutually exclusive with cidrBlock.",
+                    "type": "string"
+                },
+                "cidrBlock": {
+                    "description": "Whitelist entry in Classless Inter-Domain Routing (CIDR) notation. Mutually exclusive with ipAddress.",
+                    "type": "string"
+                },
+                "projectId": {
+                    "description": "Whitelist entry in Classless Inter-Domain Routing (CIDR) notation. Mutually exclusive with ipAddress.",
```
shum: Description is wrong

ip-whitelist/cmd/resource/resource.go
themantissa: Good call - Yes, definitely need that for this work!

ip-whitelist/cmd/resource/resource.go
themantissa: We ran into this first with Terraform.  Basically update depends on the underlying server (Atlas) to figure out what is new and what isn't so any external tool that also needs to track state may not be able to do so within that tools construct.   It's important to ensure we limit the scope of this  - in Terraform we are breaking up the IP whitelist resource to be individual entries and not the list. We have an internal ticket regarding the lack of a true update resource on this too @shum.

shum: Understood, thanks for the context!

ip-whitelist/cmd/resource/resource.go
thetonymaster: This hasn't been released yet right? I think I can add it if we do a release, I could just fetch master for the build but I think this is not desirable.

themantissa: Correct, not yet.  I've been waiting on asking for a release because the IP Whitelist resource change will be a breaking change and the AWS security groups is included in that.  But actually thinking about it we may be able to cut a release for the go client sooner since the Terraform provider should be pinned to a previous version.

ip-whitelist/cmd/resource/resource.go
thetonymaster: It is just to keep track of all entries in case one is removed/modified on the state. And we also need an ID, so we are storing it there. This is the same case with terraform.

ip-whitelist/cmd/resource/resource.go
thetonymaster: @marinsalinas can you do a release of the client? Is this ok @themantissa?

themantissa: Okay with me if okay with Marin.  I'm okay since it won't yet impact Terraform until the next release, unless someone is building from master and if so then the risk of this sort of issue should be known.

thetonymaster: I added the support for AWS Security group but I need the peering to test this

ip-whitelist/cmd/resource/resource.go
```diff
@@ -0,0 +1,218 @@
...
+	request := getProjectIPWhitelistRequest(currentModel)
+
+	_, _, err = client.ProjectIPWhitelist.Create(context.Background(), projectID, request)
+	if err != nil {
+		return handler.ProgressEvent{}, err
+	}
+
+	guid := xid.New()
+
+	currentModel.Id = encoding.NewString(guid.String())
```
shum: Should we not rely on an id returned by the call made to `client.ProjectIPWhitelist.Create` like we do in https://github.com/mongodb/mongodbatlas-cloudformation-resources/pull/7/files#diff-7da7a1a385b2ed8cb2a84a39be33b28c or is generating our own id better because we need to keep track of the whitelist seperately from Atlas?

ip-whitelist/cmd/resource/resource.go
```diff
@@ -0,0 +1,218 @@
...
+
+// Update handles the Update event from the Cloudformation service.
+func Update(req handler.Request, prevModel *Model, currentModel *Model) (handler.ProgressEvent, error) {
+	client, err := util.CreateMongoDBClient(*currentModel.ApiKeys.PublicKey.Value(), *currentModel.ApiKeys.PrivateKey.Value())
+	if err != nil {
+		return handler.ProgressEvent{}, err
+	}
+
+	err = deleteEntries(currentModel, client)
+
```
shum: delete newline

ip-whitelist/cmd/resource/resource.go
```diff
@@ -0,0 +1,218 @@
...
+	client, err := util.CreateMongoDBClient(*currentModel.ApiKeys.PublicKey.Value(), *currentModel.ApiKeys.PrivateKey.Value())
+	if err != nil {
+		return handler.ProgressEvent{}, err
+	}
+
+	err = deleteEntries(currentModel, client)
+	if err != nil {
+		return handler.ProgressEvent{
+			OperationStatus: handler.Failed,
+			Message:         "Update Failed",
```
shum:
```suggestion
			Message:         "Delete Failed",
```

ip-whitelist/cmd/resource/resource.go
```diff
@@ -0,0 +1,218 @@
...
+		if w.Comment != nil {
+			wl.Comment = *w.Comment.Value()
+		}
+		if w.CidrBlock != nil {
+			wl.CIDRBlock = *w.CidrBlock.Value()
+		}
+		if w.IpAddress != nil {
+			wl.IPAddress = *w.IpAddress.Value()
+		}
+
```
shum: delete nl

ip-whitelist/cmd/resource/resource.go
```diff
@@ -0,0 +1,218 @@
...
+		return handler.ProgressEvent{}, err
+	}
+
+	projectID := *currentModel.ProjectId.Value()
+	request := getProjectIPWhitelistRequest(currentModel)
+
+	_, _, err = client.ProjectIPWhitelist.Create(context.Background(), projectID, request)
+	if err != nil {
+		return handler.ProgressEvent{}, err
+	}
```
shum: Looks like we can replace this with the `createEntries` helper method?

ip-whitelist/cmd/resource/resource.go
```diff
@@ -0,0 +1,218 @@
...
+		}, err
+	}
+
+	request := getProjectIPWhitelistRequest(currentModel)
+	projectID := *currentModel.ProjectId.Value()
+
+	_, _, err = client.ProjectIPWhitelist.Create(context.Background(), projectID, request)
+	if err != nil {
+		return handler.ProgressEvent{}, err
+	}
```
shum: ditto above

ip-whitelist/mongodb-atlas-projectipwhitelist.json
```diff
@@ -0,0 +1,110 @@
...
+    "definitions": {
+        "whitelistDefinition": {
+            "type": "object",
+            "properties": {
+              "comment": {
+                    "description": "Comment associated with the whitelist entry.",
+                    "type": "string"
+                },
+                "ipAddress": {
+                    "description": "Whitelisted IP address. Mutually exclusive with cidrBlock.",
```
shum:
```suggestion
                    "description": "Whitelisted IP address. Mutually exclusive with cidrBlock and awsSecurityGroup.",
```

ip-whitelist/mongodb-atlas-projectipwhitelist.json
```diff
@@ -0,0 +1,110 @@
...
+                    "type": "string"
+                },
+                "cidrBlock": {
+                    "description": "Whitelist entry in Classless Inter-Domain Routing (CIDR) notation. Mutually exclusive with ipAddress.",
```
shum:
```suggestion
                    "description": "Whitelist entry in Classless Inter-Domain Routing (CIDR) notation. Mutually exclusive with ipAddress and awsSecurityGroup.",
```

ip-whitelist/cmd/resource/resource.go
thetonymaster: IP whitelist does not has an ID https://github.com/mongodb/go-client-mongodb-atlas/blob/master/mongodbatlas/project_ip_whitelist.go#L33

thetonymaster: I was tracking them with the ID, but I can track them with prev and current Model inside the function, since I am already deleting and recreating everything I don't need to track them with the ID I was creating. In order to avoid a super large ID once multiple whitelist entries are added I will just create one for this.

PR #14, opened 2020-03-13
README update

shum:
Added a couple notes to make getting started easier.


themantissa:
@shum  just a note I'm holding on this till we can get the binaries built and put in the repo for download.  That way we can skip the 1&2 steps as I think that's a lot for users to take on.  Once done we'll probably need to adjust this to just cover what is needed from there.  Ty!!!

shum:
@themantissa RFAL!

PR #16, opened 2020-04-14
Pipeline

elizhl:


shum:
I have a couple questions! I'm not very familiar with goreleaser so this was mostly an experiment for me. 

From the `mongodb-atlas-cloudformation` root dir
```
git checkout -b elizhl-master master
git pull https://github.com/elizhl/mongodbatlas-cloudformation-resources.git master
```

I installed goreleaser, following their [docs](https://goreleaser.com/install/), with
```
brew install goreleaser
```

Then I ran a dry run (doesn't publish) with
```
goreleaser release --skip-publish --snapshot
```

It released successfully quite easily, no errors. I'm guessing that I already had all the necessary libraries installed previously.

It looks like, based on [`binary: ./bin/...`](https://github.com/mongodb/mongodbatlas-cloudformation-resources/pull/16/files#diff-39e929907e3737ff7460d982cb6fbc60R5 ), that all the compiled binaries are put into `mongodbatlas-cloudformation-resources/dist/<resource>/bin`.

So while this accomplishes the task of using goreleaser to build all the binaries with one click, we don't make the cloudformation submission much more straightforward since we cannot run `cfn submit` in a dir that's not `cfn init`'d and doesn't have the necessary cfn property jsons. 

ie With all the binaries built in `dist`, if we then go to `mongodbatlas-cloudformation-resources/cluster` then try to
```
cfn submit -v --region us-east-1
```
we get the error
```
Validating your resource specification...
Packaging Go project
No Go executable was found.
Please run 'make' or the equivalent command in your IDE to compile and package the code.
```
It's not possible to direct `cfn submit` to use a dir that's not the current dir ([docs](https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-cli-submit.html)) 

TL;DR: I think what we want is for the binaries to be built and placed in the same directory as the Makefile. It seems like that would make this the most straightforward. Is that possible/adviseable? Or should we be somehow packaging the cloudformation properties also into `dist` dir? I also see that we're trying to tie goreleaser with Github Workflows, not sure if the above recommendation would make this more complicated as putting the built binaries into the main repo isn't a typical "release" per se.

cc @thetonymaster @themantissa

themantissa:
The goal is to end up with something similar to this: https://github.com/mongodb/mongocli/releases basically the idea was to  use goreleaser to easily release and make them available as downloadable binaries - then the user skips the entire build process and can go straight to the cfn part.  Hope that helps clarify what the intent is.

thetonymaster:
> LGTM. Dry run locally works great. Have not tested the Github Workflow, not sure if it's possible to do a dry run of that first.

We would also need to know what variables need to be configured for the workflow, @elizhl can you let us know so we can generate the token and everything else needed?

elizhl:
You need a secret called P_TOKEN for a token with tag and repo permissions.

PR #76, opened 2021-02-18
Update mongodb-atlas-project.json

mertsaygi:
When we try to deploy project on our AWS account, the stack has been failed with "Properties" not found error message for project. The problem is "properties" and "orgid" fields are mandatory on project creation and "properties" does not exists in property list. We can use "name" instead of that.

### DIFF
--- a/cfn-resources/project/mongodb-atlas-project.json
+++ b/cfn-resources/project/mongodb-atlas-project.json
@@ -45,7 +45,7 @@
     },
     "additionalProperties": false,
     "required": [
-        "Properties",
+        "Name",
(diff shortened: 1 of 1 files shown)


jasonmimick (approved):
Hi- 
We've had some delays with this project, and will have some accouncements coming soon.
In the meantime, this PR is fine, but the codebase will have a major update soon from the quickstart branch and more over the next 4-6 weeks.

If you'd like a sneak-peek at everything, check out: [get-started-aws-cfn](https://github.com/mongodb-developer/get-started-aws-cfn)

PR #399, opened 2023-03-07
Regen cloud-backup-schedule, snapshot, export

SuperMohit:
## Description

Please include a summary of the fix/feature/change, including any relevant motivation and context.

Link to any related issue(s): 

## Type of change:

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

## Required Checklist:

- [x] I have signed the [MongoDB CLA](https://www.mongodb.com/legal/contributor-agreement)
- [ ] I have added tests that prove my fix is effective or that my feature works
- [x] I have added any necessary documentation (if appropriate)
- [ ] I have run `make fmt` and formatted my code
- [ ] For CFN Resources: I have released by changes in the private registry and proved by change works in Atlas

PR #401, opened 2023-03-07
Regen Updated global-cluster, maint wind, ol9-arch, org-invite

SuperMohit:
## Description

Please include a summary of the fix/feature/change, including any relevant motivation and context.

Link to any related issue(s): 

## Type of change:

- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] This change requires a documentation update

## Required Checklist:

- [ ] For CFN Resources: I have released by changes in the private registry and proved by change works in Atlas

### DIFF
--- a/cdk/cdk-resources/global-cluster-config/API.md
+++ b/cdk/cdk-resources/global-cluster-config/API.md
@@ -403,7 +403,7 @@ Check whether the given construct is a CfnResource.
 | <code><a href="#@mongodbatlas-awscdk/global-cluster-config.CfnGlobalClusterConfig.property.ref">ref</a></code> | <code>string</code> | Return a string that will be resolved to a CloudFormation `{ Ref }` for this element. |
 | <code><a href="#@mongodbatlas-awscdk/global-cluster-config.CfnGlobalClusterConfig.property.cfnOptions">cfnOptions</a></code> | <code>aws-cdk-lib.ICfnResourceOptions</code> | Options for this resource, such as condition, update policy etc. |
 | <code><a href="#@mongodbatlas-awscdk/global-cluster-config.CfnGlobalClusterConfig.property.cfnResourceType">cfnResourceType</a></code> | <code>string</code> | AWS resource type. |
-| <code><a href="#@mongodbatlas-awscdk/global-cluster-config.CfnGlobalClusterConfig.property.attrCustomZoneMappings">attrCustomZoneMappings</a></code> | <code>any[]</code> | Attribute `MongoDB::Atlas::GlobalClusterConfig.CustomZoneMappings`. |
+| <code><a href="#@mongodbatlas-awscdk/global-cluster-config.CfnGlobalClusterConfig.property.attrRemoveAllZoneMapping">attrRemoveAllZoneMapping</a></code> | <code>aws-cdk-lib.IResolvable</code> | Attribute `MongoDB::Atlas::GlobalClusterConfig.RemoveAllZoneMapping`. |
 | <code><a href="#@mongodbatlas-awscdk/global-cluster-config.CfnGlobalClusterConfig.property.props">props</a></code> | <code><a href="#@mongodbatlas-awscdk/global-cluster-config.CfnGlobalClusterConfigProps">CfnGlobalClusterConfigProps</a></code> | Resource props. |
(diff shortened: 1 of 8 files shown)


SuperMohit:
Warning:
generating for MongoDB::Atlas::GlobalClusterConfig
Unsupported type {"type":"array","insertionOrder":false,"description":"List that contains comma-separated key value pairs to map zones to geographic regions. These pairs map an ISO 3166-1a2 location code, with an ISO 3166-2 subdivision code when possible, to the human-readable label for the desired custom zone. MongoDB Cloud maps the ISO 3166-1a2 code to the nearest geographical zone by default. Include this parameter to override the default mappings.\n\nThis parameter returns an empty object if no custom zones exist.","items":{"$ref":"#/definitions/zoneMapping","type":"object"}} for read-only property (attribute) MongoDB::Atlas::GlobalClusterConfig.CustomZoneMappings

maastha:
> Warning: generating for MongoDB::Atlas::GlobalClusterConfig Unsupported type {"type":"array","insertionOrder":false,"description":"List that contains comma-separated key value pairs to map zones to geographic regions. These pairs map an ISO 3166-1a2 location code, with an ISO 3166-2 subdivision code when possible, to the human-readable label for the desired custom zone. MongoDB Cloud maps the ISO 3166-1a2 code to the nearest geographical zone by default. Include this parameter to override the default mappings.\n\nThis parameter returns an empty object if no custom zones exist.","items":{"$ref":"#/definitions/zoneMapping","type":"object"}} for read-only property (attribute) MongoDB::Atlas::GlobalClusterConfig.CustomZoneMappings


@SuperMohit Could you please elaborate more on this warning? 
Is there a fix we should take up? 
Is there any impact on UX?

SuperMohit:
@maastha - @Adelmar92 is looking into this for other resource. It's the similar warning.

Adelmar92:
@SuperMohit @maastha it seems that that resource is trying to output a list of objects, we had a meeting with Brian from AWS, it seems that cloud formation does not allow to define array of objects as "readonly" properties, but you have no warning on the CFN about that, but you do on the CDK, so we should remove that property as ReadOnly

SuperMohit:
> zoneMapping



> @SuperMohit @maastha it seems that that resource is trying to output a list of objects, we had a meeting with Brian from AWS, it seems that cloud formation does not allow to output array of objects as "readonly" properties, but you have no warning on the CFN about that, but you do on the CDK, so we should remove that property as ReadOnly

Thanks @Adelmar92. When there is this warning, it is skipped already.
https://github.com/cdklabs/cdk-import/blob/3c0b7f039df581eb3e236bf0dd3f658793cad9ec/src/cfn-resource-generator.ts#L42

maastha:
> 

So you're saying we need to make these changes to CFN first?

Adelmar92:
@maastha @SuperMohit , so the thing is we should remove those as read only, but is not mandatory, its just a warning, we can publish the CDK and update them later, is not going to prevent users from using the CDK, the will not be able to use the property only. so we can decide, update the CFN and republish everything now, or we can create an issue and update them later

maastha:
> @maastha @SuperMohit , so the thing is we should remove those as read only, but is not mandatory, its just a warning, we can publish the CDK and update them later, is not going to prevent users from using the CDK, the will not be able to use the property only. so we can decide, update the CFN and republish everything now, or we can create an issue and update them later

Can you clarify what you mean by "the will not be able to use the property only" ? What is the user impact here? I don't think it's a blocker but let's test all cases manually since it's a number of resources and then @Zuhairahmed @colm-quinn can take a call.

SuperMohit:
> #399

@maastha Thanks for your inputs. But, I am not getting what type of manual testing is required? CFN doesn't allow Outputs which are array of an Object type(a limitation on AWS side). This is the same warning CDK is throwing. If we have to fix this, we need to remove all the read-only properties which are of type object from CFN schema.

maastha (commented):
LGTM, please share the outputs of the stack and create a JIRA ticket to address warnings.

maastha (approved):
Warning will be addressed as part of [INTMDB-670](https://jira.mongodb.org/browse/INTMDB-670)

PR #443, opened 2023-03-22
INTMDB-681: fixed cluster label

Adelmar92:
## Description

Fixed Label for Cluster resource:

- Changed the Infrastructure Tool Label from _MongoDB Atlas Terraform Provider_ to _MongoDB Atlas CloudFormation Provider_
- Refactor logic to automatically add the label once the resource is created or updated

Link to any related issue(s): 

## Type of change:

- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)

## Required Checklist:

- [ ] For CFN Resources: I have released by changes in the private registry and proved by change works in Atlas


cfn-resources/cluster/cmd/resource/resource.go
```diff
@@ -1061,3 +1044,15 @@ func setClusterRequest(currentModel *Model, err error) (*mongodbatlas.AdvancedCl
 	clusterRequest.TerminationProtectionEnabled = currentModel.TerminationProtectionEnabled
 	return clusterRequest, handler.ProgressEvent{}, nil
 }
+
+func (m *Model) validateDefaultLabel() *handler.ProgressEvent {
+	if containsLabelOrKey(m.Labels, defaultLabel) {
+		_, _ = log.Warnf("Update - error :%s", LabelError)
+		pe := progress_events.GetFailedEventByCode(LabelError, cloudformation.HandlerErrorCodeInvalidRequest)
+		return &pe
+	}
```
andreaangiolillo, line 1053: [q] Why do you fail if it has already the label?

cfn-resources/cluster/cmd/resource/resource.go
```diff
@@ -197,6 +201,11 @@ func Update(req handler.Request, prevModel *Model, currentModel *Model) (handler
 		return updateClusterCallback(client, currentModel, *currentModel.ProjectId)
 	}
 
+	pe := currentModel.validateDefaultLabel()
+	if pe != nil {
+		return *pe, nil
+	}
+
```
andreaangiolillo, line 208: if the cluster has been created with CFN, it will have the label and then `validateDefaultLabel` will fail. I believe we want to add the label only for cluster creation not update @Zuhairahmed to confirm

Adelmar92: let me try that, because if its coming from the current model, the user should not be providing the tag, ill test that just in case

Adelmar92: so this is the case, if the user does not provide the label in the template used as an input on the update is fine, if the user provides the label as the update imput, it will brake and the user will need to provide a new template without the label

Zuhairahmed: @andreaangiolillo is correct. @Adelmar92 if not possible to improve further can you make ^^ this clear in Readme file for this resource.

Zuhairahmed (approved):

Zuhairahmed (approved):
LGTM pending APIx review

cfn-resources/cluster/cmd/resource/resource.go
andreaangiolillo: yes, we should not return an error if the label is already there

cfn-resources/cluster/cmd/resource/resource.go
andreaangiolillo: > so this is the case, if the user does not provide the label in the template used as an input on the update is fine, if the user provides the label as the update imput, it will brake and the user will need to provide a new template without the label

isn't it because of `validateDefaultLabel`? This function returns an error if the model has a label (which is wrong).  The idea is that users can add as many labels as they want. We should add the `Infrastructure Tool` all the time that the cluster is created with CFN. As a result, ]

- if the user does not provide a label, we add `Infrastructure Tool`
- if the user provides 4 labels, we add 4 labels + `Infrastructure Tool`

Let me know if I am missing something or if it is not clear. I know that it can be confusing. Thanks!

andreaangiolillo (commented):
Left a few comments

cfn-resources/cluster/cmd/resource/resource.go
andreaangiolillo: My understanding is that we would replace it. @Zuhairahmed to confirm here

Adelmar92: so we discuss this with @Zuhairahmed im modifying the behaviour like this:

- if the user provides the label Infrastructure Tool we are keepeng it
- if the user does not provides the label Infrastructure Tool we are adding it

andreaangiolillo (approved):
LGTM

PR #560, opened 2023-05-15
Return connectionId when creating new peering connection

swachter:
The connectionId should be output because it may be fed into a resource for accepting the connection.

### DIFF
--- a/cfn-resources/network-peering/cmd/resource/resource.go
+++ b/cfn-resources/network-peering/cmd/resource/resource.go
@@ -108,6 +108,7 @@ func Create(req handler.Request, prevModel *Model, currentModel *Model) (handler
 
 	_, _ = logger.Debugf("Create peerResponse:%+v", peerResponse)
 	currentModel.Id = &peerResponse.ID
+	currentModel.ConnectionId = &peerResponse.ConnectionID
 
 	return handler.ProgressEvent{
 		OperationStatus: handler.Success,


Zuhairahmed:
@Adelmar92 can you review? if no concerns feel free to approve and merge. Thanks for PR [swachter](https://github.com/swachter)!

swachter:
> if the user might require to output the ConnectionId property, it needs to be added in the readOnlyProperties definition in mongodb-atlas-networkpeering.json, also needs to ensure that the property is returned in the Read function

The ConnectionId is already included in the readOnlyProperties. It is returned as part of the [createPeeringConnection](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/#tag/Network-Peering/operation/createPeeringConnection) ressponse.

Adelmar92:
Hello @swachter if that is the case i think assigning the property on the create wont have any effect, you can still update the property in the current resource. here is an example

here is the output in the stack: 
image

and the output: 
image

is this what you are mentioning? 
if not can you explain me how would the user access the property if its returned in the CREATE stage?

thanks!!!

StefanWachter1507:
I assume that in your example ConnId is `null` when the stack is deployed the first time, i.e. a new peering connection is created.

Here is my use case: I try to use the `networkPeering.ConnectionId` to establish routes in private subnets. The CDK code looks like:

```
        const peering = new ma.CfnNetworkPeering(this, 'networkPeering', {
            awsAccountId: props.env.account,
            accepterRegionName: props.env.region,
            projectId,
            vpcId: props.vpc.vpcId,
            containerId: props.networkContainer.attrId,
            routeTableCidrBlock: props.vpc.vpcCidrBlock
        });
        ...
        // -> the deployment of the route constructs fails because peering.attrConnectionId is null
        // -> if the route constructs are include in the stack based on a condition and the condition is false on the first deployment
        //    and true on the second deployment then the routes get deployed
        // -> the CfnNetworkPeering construct is able to read the ConnectionId of an existing peering connection
        //    but does not provide the ConnectionId when a new peering connection is created 
        props.vpc.privateSubnets.forEach(psn => {
            new CfnRoute(this, `privateSubnetPeering${psn.availabilityZone}`, {
                destinationCidrBlock: props.atlasCidrBlock,
                routeTableId: psn.routeTable.routeTableId,
                vpcPeeringConnectionId: peering.attrConnectionId
            });
        });

```

The CloudFormation template references the `ConnectionId` like you showed using the `Fn:GetAtt` function. The snippet for a route resource looks like:

```
  "privateSubnetPeeringeucentral1a": {
   "Type": "AWS::EC2::Route",
   "Properties": {
    "RouteTableId": {
     "Fn::ImportValue": "payment-develop-vpc:ExportsOutputRefVpcPrivateSubnet1RouteTableB2C5B500BF514E65"
    },
    "DestinationCidrBlock": "192.168.216.0/21",
    "VpcPeeringConnectionId": {
     "Fn::GetAtt": [
      "networkPeering",
      "ConnectionId"
     ]
    }
   }
  }
```

Adelmar92:
> I assume that in your example ConnId is `null` when the stack is deployed the first time, i.e. a new peering connection is created.
> 
> Here is my use case: I try to use the `networkPeering.ConnectionId` to establish routes in private subnets. The CDK code looks like:
> 
> ```
>         const peering = new ma.CfnNetworkPeering(this, 'networkPeering', {
>             awsAccountId: props.env.account,
>             accepterRegionName: props.env.region,
>             projectId,
>             vpcId: props.vpc.vpcId,
>             containerId: props.networkContainer.attrId,
>             routeTableCidrBlock: props.vpc.vpcCidrBlock
>         });
>         ...
>         // -> the deployment of the route constructs fails because peering.attrConnectionId is null
>         // -> if the route constructs are include in the stack based on a condition and the condition is false on the first deployment
>         //    and true on the second deployment then the routes get deployed
>         // -> the CfnNetworkPeering construct is able to read the ConnectionId of an existing peering connection
>         //    but does not provide the ConnectionId when a new peering connection is created 
>         props.vpc.privateSubnets.forEach(psn => {
>             new CfnRoute(this, `privateSubnetPeering${psn.availabilityZone}`, {
>                 destinationCidrBlock: props.atlasCidrBlock,
>                 routeTableId: psn.routeTable.routeTableId,
>                 vpcPeeringConnectionId: peering.attrConnectionId
>             });
>         });
> ```
> 
> The CloudFormation template references the `ConnectionId` like you showed using the `Fn:GetAtt` function. The snippet for a route resource looks like:
> 
> ```
>   "privateSubnetPeeringeucentral1a": {
>    "Type": "AWS::EC2::Route",
>    "Properties": {
>     "RouteTableId": {
>      "Fn::ImportValue": "payment-develop-vpc:ExportsOutputRefVpcPrivateSubnet1RouteTableB2C5B500BF514E65"
>     },
>     "DestinationCidrBlock": "192.168.216.0/21",
>     "VpcPeeringConnectionId": {
>      "Fn::GetAtt": [
>       "networkPeering",
>       "ConnectionId"
>      ]
>     }
>    }
>   }
> ```
> 
> I think, the only piece missing is to copy the `ConnectionId` from the create (POST) response back into the model.

nono in my example connId gets completed.
In your case when you do "Fn::GetAtt": ["networkPeering", "ConnectionId"] the ConnectionId is null? because the behaviour with the GetAtt is that is reading everything coming out of the READ function not the Create

Adelmar92:
@StefanWachter1507 did you add dependencies between the resources? since AWS::EC2::Route is using a network peering property you might want to add them as dependent  using the DependsOn tag like this:

``` json
  "privateSubnetPeeringeucentral1a": {
   "Type": "AWS::EC2::Route",
   "DependsOn" : "networkPeering"
   "Properties": {
    "RouteTableId": {
     "Fn::ImportValue": "payment-develop-vpc:ExportsOutputRefVpcPrivateSubnet1RouteTableB2C5B500BF514E65"
    },
    "DestinationCidrBlock": "192.168.216.0/21",
    "VpcPeeringConnectionId": {
     "Fn::GetAtt": [
      "networkPeering",
      "ConnectionId"
     ]
    }
   }
  }
```

StefanWachter1507:
@Adelmar92 I added the dependency as you suggested. Unfortunately it does not work. When deploying the stack I get:
```
 Attribute 'ConnectionId' does not exist
```
The snippet with the route resource now look the following:

```
"privateSubnetPeeringeucentral1a": {
   "Type": "AWS::EC2::Route",
   "Properties": {
    "RouteTableId": {
     "Fn::ImportValue": "payment-develop-vpc:ExportsOutputRefVpcPrivateSubnet1RouteTableB2C5B500BF514E65"
    },
    "DestinationCidrBlock": "192.168.216.0/21",
    "VpcPeeringConnectionId": {
     "Fn::GetAtt": [
      "networkPeering",
      "ConnectionId"
     ]
    }
   },
   "DependsOn": [
    "networkPeering"
   ]
  }
```

StefanWachter1507:
I use the `MongoDB::Atlas::Project` resource simililarly. There I copy the `projectId` of a newly created project into the SSM parameter store:
```
        this.project = new ma.CfnProject(this, 'project', {
            name: projectName,
            orgId: MONGO_ATLAS_ORG_ID
        });

        new StringParameter(this, 'projectId', {
            parameterName: mongoProjectIdParamName(props) + '2',
            description: `Project (aka Group) ID of Mongo Atlas project for ${props.vertical}`,
            dataType: ParameterDataType.TEXT,
            stringValue: this.project.attrId
        });
```

When I look into the `create` function in the Go code the project resource, I see that the `projectId` is copied from the create (POST) response (cf. [project/cmd/resource/resource.go#L125](https://github.com/mongodb/mongodbatlas-cloudformation-resources/blob/master/cfn-resources/project/cmd/resource/resource.go#L125)). I expected that the same mechanism would work for the `connectionId` that is returned in the create (POST) response of triggers.

Adelmar92:
hi @StefanWachter1507 i was able to replicate the error, the problem is somewhere else
i have 2 possible scenarios
problem 1:  I found that the resource does not wait or validate if the network peering is created before returning success, so posibly the resource returns Create while is creating, so you dont have the connectionId property yet.
problem 2: The connection ID is not set if the network peering is fails. 

ill try to solve the problem 1, can you please check the problem 2, see if the resource is provided with all the parameters, and they are right, you can try creating the resource manually directly on cloud formation

Adelmar92:
@StefanWachter1507 ive created this pull request https://github.com/mongodb/mongodbatlas-cloudformation-resources/pull/602 to fix the issue, i need to wait until is approved to publish it if we are lucky we can get it done tomorrow, can you please tell me in wich region you are testing, so i can publsh a beta version to see if that fixes your issue?

StefanWachter1507:
@Adelmar92 Many thanks for digging into the issue and providing a fix!!! I work in the region "eu-central-1".

Adelmar92:
@StefanWachter1507 a version is been deployed to eu-central-1 im working on the other regions now, can you give it a try? remember that now if the networkpeering connection ends in  FAILED status, the stack will fail too

StefanWachter1507:
Works like a charm! Many thanks!!!

PR #744, opened 2023-10-02
feat: migrate third-party resource to use the autogenerated Atlas SDK

andreaangiolillo:
## Proposed changes

_Jira ticket:_ [INTMDB-1108](https://jira.mongodb.org/browse/INTMDB-1108)

Please include a summary of the fix/feature/change, including any relevant motivation and context.

Link to any related issue(s): 

## Type of change:

  expected)

## Manual QA performed:

- [x] Validated in Atlas UI
- [x] Included screenshots

## Required Checklist:

  works in Atlas

### CFN test
```bash
cfn test --function-name TestEntrypoint --verbose

=== 13 passed, 2 skipped, 9 deselected in 72.36s (0:01:12) ===
```

### CFN Stack
Screenshot 2023-10-02 at 17 03 46
Screenshot 2023-10-02 at 17 02 46

### DIFF
--- a/cfn-resources/third-party-integration/cmd/resource/resource.go
+++ b/cfn-resources/third-party-integration/cmd/resource/resource.go
@@ -264,8 +264,8 @@ func List(req handler.Request, prevModel *Model, currentModel *Model) (handler.P
 	}
 
 	mm := make([]interface{}, 0)
-	for _, integration := range integrations.Results {
-		m := integrationToModel(*currentModel, &integration)
+	for i := range integrations.Results {
+		m := integrationToModel(*currentModel, &integrations.Results[i])
 		mm = append(mm, m)
 	}
 
--- a/cfn-resources/third-party-integration/cmd/resource/resource.go
+++ b/cfn-resources/third-party-integration/cmd/resource/resource.go
@@ -16,19 +16,16 @@ package resource
 
 import (
 	"context"
-	"fmt"
 	"net/http"
 
 	"github.com/aws-cloudformation/cloudformation-cli-go-plugin/cfn/handler"
-	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/cloudformation"
-	"github.com/mongodb/mongodbatlas-cloudformation-resources/profile"
 	"github.com/mongodb/mongodbatlas-cloudformation-resources/util"
 	"github.com/mongodb/mongodbatlas-cloudformation-resources/util/constants"
 	log "github.com/mongodb/mongodbatlas-cloudformation-resources/util/logger"
-	progressevents "github.com/mongodb/mongodbatlas-cloudformation-resources/util/progressevent"
+	"github.com/mongodb/mongodbatlas-cloudformation-resources/util/progressevent"
 	"github.com/mongodb/mongodbatlas-cloudformation-resources/util/validator"
-	"go.mongodb.org/atlas/mongodbatlas"
+	"go.mongodb.org/atlas-sdk/v20230201008/admin"
 )
 
 var RequiredFields = []string{constants.IntegrationType, constants.ProjectID}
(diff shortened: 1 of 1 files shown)


marcosuma (dismissed):

cfn-resources/third-party-integration/cmd/resource/resource.go
```diff
@@ -80,65 +73,51 @@ func Create(req handler.Request, prevModel *Model, currentModel *Model) (handler
 
 	// checking per type fields
 	requiredInputs := requiredPerType[*IntegrationType]
-	// Validation
-	IntegrationTypeValidation := validateModel(requiredInputs, currentModel)
-	if IntegrationTypeValidation != nil {
-		return *IntegrationTypeValidation, nil
+	validationErr := validateModel(requiredInputs, currentModel)
+	if validationErr != nil {
```
lantoli, line 77: knitpit:    if validationErr := validateModel(requiredInputs, currentModel); validationErr != nil {

cfn-resources/third-party-integration/cmd/resource/resource.go
```diff
@@ -150,132 +129,109 @@ func Update(req handler.Request, prevModel *Model, currentModel *Model) (handler
...
-		ResourceModel:   integrationToModel(*currentModel, integrations.Results[0]),
+		ResourceModel:   integrationToModel(*currentModel, &integrations.Results[0]),
 	}, nil
 }
 
-func updateIntegrationFromSchema(currentModel *Model, integration *mongodbatlas.ThirdPartyIntegration) {
-	if currentModel.Url != nil && *currentModel.Url != integration.URL {
-		integration.URL = *currentModel.Url
+func updateIntegrationFromSchema(currentModel *Model, integration *admin.ThridPartyIntegration) {
+	if currentModel.Url != nil && *currentModel.Url != *integration.Url {
```
lantoli, line 163: I think currentModel.Url is *string, do we want to assign if *currentModel.Url == "", the if clause will be true. 
same for all conditions here.

in case we don't want to assign if it is empty, we can do something like this: 
if util.IsStringPresent(currentModel.Url) && !util.AreStringPtrEqual(currentModel.Url, integration.Url) {
although to be honest i don't think it makes any harm to assign even if it's the same, we could just have: 

if  util.IsStringPresent(currentModel.Url) {
	integration.Url = currentModel.Url
}

lantoli: I see you're doing my last idea in fucntion modelToIntegration, why not here?

lantoli (dismissed):

cfn-resources/third-party-integration/cmd/resource/resource.go
AgustinBettati: Both options are okay with me, mainly want to avoid the `*integration.Url` as it has the risk of nil pointer dereference. (Applies to other optional properties below in this function)

lantoli: I agree that they behave similarly, although I find easier to understand and cleaner: if util.IsStringPresent(currentModel.Url) {

andreaangiolillo: > I see you're doing my last idea in fucntion modelToIntegration, why not here?
the only reason is that `modelToIntegration` was only checking for empty string whereas `updateIntegrationFromSchema` check that the current model was different from the input

cfn-resources/third-party-integration/cmd/resource/resource.go
andreaangiolillo: updated

AgustinBettati (approved):
LGTM

cfn-resources/third-party-integration/cmd/resource/resource.go
lantoli: ok, but I don't understand yet why we use !util.AreStringPtrEqual in one method and not in the other, what's the different logic we want to achieve?

andreaangiolillo: `modelToIntegration` is used for the CREATE so we know there is no resource in Atlas. `updateIntegrationFromSchema` is used in the UPDATE, so we want to send in the body request only the fields that have been changed in the currentModel. The main point here is that we have fields such as API tokens that we don't want to send again into the API call if not really needed.

PR #759, opened 2023-10-06
feat: HELP-50727 cloud buckup restore job already exist validation

Adelmar92:
## Proposed changes

**FIXED READ AND DELETE VALIDATION FOR ALREADY EXISTS ERRORS**

CloudFormation requires any third-party resource to fail with a "NotFound" error in case the user tries to read it after being deleted.

In order to comply with this behavior, we introduce the following behavior on the resource:

- **If the job is marked as Cancelled:** It will fail with a Not Found error, both in the READ and DELETE flows.
- **if the Job is marked as Finished, Expired or Failed:** The READ will be executed (not returning an error), 
and the DELETE flow will return success, being these final states.

added changelog for user guidance

_Jira ticket:_ [HELP-50727](https://jira.mongodb.org/browse/HELP-50727)

## Type of change:

  expected)

## Manual QA performed:

- [ ] Validated in Atlas UI
- [ ] Included screenshots

## Required Checklist:

  works in Atlas


cfn-resources/cloud-backup-restore-jobs/cmd/resource/resource.go
```diff
@@ -482,10 +487,11 @@ func getRestoreJob(client *mongodbatlas.Client, currentModel *Model) (*mongodbat
...
 
 func isJobFinished(job mongodbatlas.CloudProviderSnapshotRestoreJob) bool {
-	failed := job.Failed != nil && *job.Failed
-	finished := job.FinishedAt != ""
+	return job.FinishedAt != ""
+}
 
-	return failed || job.Cancelled || job.Expired || finished
+func isJobFailed(job mongodbatlas.CloudProviderSnapshotRestoreJob) bool {
+	return job.Failed != nil && *job.Failed
```
lantoli, line 494: why don't we have Failed in the model?  I think it would be easier, e.g. we wouldn't need isJobFailed to use the API object and could just use the model

examples/cloud-backup-restore-jobs/restore.json
```diff
@@ -40,7 +45,7 @@
           "Ref": "ProjectId"
         },
         "ClusterName": {
```
lantoli, line 47: are we keeping ClusterName?

lantoli (dismissed):

cfn-resources/cloud-backup-restore-jobs/CHANGELOG.md
```diff
@@ -0,0 +1,75 @@
...
+
+example:
+
+Original:
+``` yaml
+RestoreJob:
+Type: MongoDB::Atlas::CloudBackUpRestoreJobs
+Properties:
+ProjectId: !Ref 'ProjectSource'
+ClusterName: !Ref 'ClusterNameSource' <----- Old identifier (deprecated)
```
AgustinBettati, line 24: This field has been completely removed right? In this case would modify `deprecated` with `removed`

cfn-resources/cloud-backup-restore-jobs/cmd/resource/resource.go
Adelmar92: apart of adding this property, we have to think if is going to be a valid property for the user, if is going to be used only here i dont see a reason to add it to the model

examples/cloud-backup-restore-jobs/restore.json
Adelmar92: this is for the snapshot, it has not been changed yet

PR #782, opened 2023-10-12
fix!: INTMDB-1170 searchindex mapping issue

Adelmar92:
## Proposed changes

Originally, the mapping fields were inputted like this:
``` json
mappings: {
fields: [
"ownershipDetails.assetName:string"
],
},
``` 

the problem with this approach  is that the user will only be able to input one field and the time, and also only the type
was specified, some data types like "autocomplete" may require extra parameters. (like analyzer, maxGrams, minGrams, etc),
also a field would have accepted only one data type, preventing the user to setting multiple types to a single field. 

we required a way to input dynamic structures to each property, but since cloud formation does not provide a way to
input dynamic objects as parameters, we decided to input the data type as a stringify json,
representing a list of types like is shown in the MongoDB atlas UI

this is has been discussed with @Zuhairahmed we decided to increase the mayor version and implement the definitive solution. since this will not impact a grate amount of users

Original:
``` json
mappings: {
fields: [
"property_type:string"
],
},
``` 

New:
``` json
mappings: {
fields:"{\"property_type\": {\"type\": \"string\"}}",
},
```

tests:

this template was created with the new field format

image

result:

image

Update test:
to simulate the update to this new changes, i created a stack with the old version:

image

the creation was completed
then ive submitted the new version, and uploaded a new stack, with the new mapping format:
the change was detected as Replacement:False
image
image
image

the update was completed successfully 

Jira issue: [INTMDB-1170](https://jira.mongodb.org/browse/INTMDB-1170)

## Type of change:

  expected)

## Manual QA performed:

- [x] Validated in Atlas UI
- [x] Included screenshots

## Required Checklist:

  works in Atlas


cfn-resources/search-index/cmd/resource/resource.go
```diff
@@ -205,18 +206,31 @@ func newMappingsFields(fields []string) (map[string]interface{}, error) {
 
 	fieldsMap := make(map[string]interface{})
 	for _, p := range fields {
-		f := strings.Split(p, ":")
+		f := strings.SplitN(p, ":", 2)
```
marcosuma, line 209: what was the reason for changing this?

cfn-resources/search-index/CHANGELOG.md
```diff
@@ -0,0 +1,49 @@
+# Changelog
+
+## (2023-10-12)
```
marcosuma, line 3: Is this a breaking change? I think so?

cfn-resources/search-index/cmd/resource/resource.go
```diff
@@ -205,18 +206,31 @@ func newMappingsFields(fields []string) (map[string]interface{}, error) {
...
 		}
-		fieldsMap[f[0]] = map[string]interface{}{
-			"type": f[1],
+
+		fieldType, err := parseFieldType(f[1])
+		if err != nil {
+			return nil, err
 		}
+
+		fieldsMap[f[0]] = fieldType
```
marcosuma, line 219: curious: what has been the testing strategy here? Is the only available test the one you did manually in the PR description?

cfn-resources/search-index/CHANGELOG.md
```diff
@@ -0,0 +1,49 @@
...
+"property_type:string"
+],
+},
+``` 
+
+New:
+``` json
+mappings: {
+fields: [
+"property_type:[{\"type\":\"autocomplete\"},{\"type\":\"string\"}}]"
```
AgustinBettati, line 37: I find it confusing that the string value has a mix between non json `property_type:` and then a proper json format `[{\"type\":\"autocomplete\"},{\"type\":\"string\"}}]`.
Looking into how we define `mappings_fields` in terraform, it has a single property that expects a json format in its entirety to define all mappings: https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/resources/search_index#mappings_fields
I think this will be more clear to the end user, what do you think?

examples/search-index/searchIndexStaticFields.json
```diff
@@ -0,0 +1,65 @@
...
+        "ProjectId":{
+          "Ref":"ProjectId"
+        },
+        "Name":{
+          "Ref":"IndexName"
+        },
+        "Mappings":{
+          "Dynamic":false,
+          "Fields":[
+            "property_type:[{\"type\":\"autocomplete\"},{\"type\":\"string\"}]"
```
AgustinBettati, line 57: apart from adjusting the example the documentation for the `Fields` property has not been updated.

cfn-resources/search-index/CHANGELOG.md
AgustinBettati: Yes this would be breaking change. We could analyse deprecating (and not removing) the existing "Fields" property, and add a new one which expects the new json format.

examples/search-index/searchIndexStaticFields.json
```diff
@@ -0,0 +1,65 @@
...
+          ]
+        },
+        "SearchAnalyzer":"lucene.standard"
+      }
+    }
+  }
+}
```
AgustinBettati, line 64: Would include the new format in contract testing

AgustinBettati (changes requested):

cfn-resources/search-index/CHANGELOG.md
Adelmar92: so this is my original approach, the mappings is a json file, the problem is that if i implement it like this, im not able to update an existing stack. for example:
if an user is using search index with a list of strings (asi it is right now), when he tries to use the new version, and execute the update, this will throw an error, because cfn will try to populate the previous model with an interface{} and will fail

cfn-resources/search-index/cmd/resource/resource.go
Adelmar92: ive added a contract testing for this one

cfn-resources/search-index/cmd/resource/resource.go
Adelmar92: im only going to split in two parts, since the second one is a json may contain :

cfn-resources/search-index/CHANGELOG.md
AgustinBettati: > so this is my original approach, the mappings is a json file, the problem is that if i implement it like this, im not able to update an existing stack

Have you considered a new field which allows defining a single string with all the json configuration? And we can deprecate the old `fields` property but leave it available to avoid any breaking changes. These 2 fields could be mutually exclusive, meaning users would use one approach or the other. What do you think? I would imagine user could update the stack from using one field to another as long as the underlying content is the same.

cfn-resources/search-index/cmd/resource/resource.go
```diff
@@ -205,18 +206,31 @@ func newMappingsFields(fields []string) (map[string]interface{}, error) {
...
 	}
 
 	return fieldsMap, nil
 }
 
+func parseFieldType(jsonStr string) (interface{}, error) {
+	var result interface{}
+	err := json.Unmarshal([]byte(jsonStr), &result)
```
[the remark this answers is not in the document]
Adelmar92: yes is actually a braking change

lantoli:
closing because code is included in this merged PR: https://github.com/mongodb/mongodbatlas-cloudformation-resources/pull/801

PR #810, opened 2023-11-08
fix: Intmdb 1135 fix team users update

marcosuma:
## Proposed changes

_Jira ticket:_ https://jira.mongodb.org/browse/INTMDB-1135

Please include a summary of the fix/feature/change, including any relevant motivation and context:

at the basis of the bug this PR fixes there are two main problems:
* the logic of updating usernames in the teams is wrong. Right now [if the user specified in input is already in the team](https://github.com/mongodb/mongodbatlas-cloudformation-resources/blob/master/cfn-resources/teams/cmd/resource/resource.go#L253) it **gets removed**  (for a reason that I haven't understood). But this seems wrong to me and also confirmed as the original bug that caused this PR.
* [this method](https://github.com/mongodb/mongodbatlas-cloudformation-resources/blob/master/cfn-resources/teams/cmd/resource/resource.go#L275) was not even calling the `.Execute()` method, which practically means users were never added in `update` phase.

I added unit tests that are covering the new logic. Longer conversation is [here](https://mongodb.stackenterprise.co/questions/1061). Interesting read related to this change is ["Accept interfaces, return structs"](https://bryanftan.medium.com/accept-interfaces-return-structs-in-go-d4cab29a301b). 

On top of unit tests, I also tested manually:
* create team with username
* update team removing username
* update team readding username
* update team by changing name

Link to any related issue(s): 

## Type of change:

  expected)

## Manual QA performed:

- [ ] Validated in Atlas UI
- [ ] Included screenshots

## Required Checklist:

  works in Atlas


cfn-resources/teams/cmd/util/user_test.go
```diff
@@ -0,0 +1,62 @@
...
+	usernames := []string{"validuser1", "invaliduser1", "validuser2", "invaliduser2"}
+
+	// Set up mock expectations for successful and unsuccessful calls to GetUserByUsername
+	mockAtlasV2Client.EXPECT().GetUserByUsername(context.Background(), "validuser1").Return(&atlasv2.CloudAppUser{}, nil)
+	mockAtlasV2Client.EXPECT().GetUserByUsername(context.Background(), "invaliduser1").Return(nil, errors.New("invalid username"))
+	mockAtlasV2Client.EXPECT().GetUserByUsername(context.Background(), "validuser2").Return(&atlasv2.CloudAppUser{}, nil)
+	mockAtlasV2Client.EXPECT().GetUserByUsername(context.Background(), "invaliduser2").Return(nil, errors.New("invalid username"))
+
+	// Call the FilterOnlyValidUsernames function
+	validUsers := FilterOnlyValidUsernames(mockAtlasV2Client, usernames)
```
marcosuma, line 56: This is currently generating a syntax error:

```
cannot use mockAtlasV2Client (variable of type *mock_admin.MockMongoDBCloudUsersApi) as admin.MongoDBCloudUsersApi value in argument to FilterOnlyValidUsernames: *mock_admin.MockMongoDBCloudUsersApi does not implement admin.MongoDBCloudUsersApi (missing method createUserExecute)
```

marcosuma:
Things to discuss/improve:
1. Use `gomock` or not; if not, what alternatives? (gomock is no longer maintained but it offer this nice `EXPECT()` functionality)
2. naming: is `team_user` good enough as file name? Should we separate that file further?
3. service: should we add all endpoints being used by `teams` resource into the service? ideally that would be good, just taking a bit more refactoring and time

cfn-resources/teams/cmd/resource/team-user/team_user.go
```diff
@@ -0,0 +1,136 @@
...
+	return s.mongoDBCloudUsersAPI.GetUserByUsername(context.Background(), userName).Execute()
+}
+
+func (s *TeamUsersAPIService) AddTeamUser(ctx context.Context, orgID string, teamID string, addUserToTeam *[]atlasv2.AddUserToTeam) (*atlasv2.PaginatedApiAppUser, *http.Response, error) {
+	return s.teamsAPI.AddTeamUser(ctx, orgID, teamID, addUserToTeam).Execute()
+}
+
+func (s *TeamUsersAPIService) RemoveTeamUser(ctx context.Context, orgID string, teamID string, userID string) (*http.Response, error) {
+	return s.teamsAPI.RemoveTeamUser(ctx, orgID, teamID, userID).Execute()
+}
```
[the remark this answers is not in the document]
marcosuma: if the signature changed, wouldn't there be the same problem just in a different file?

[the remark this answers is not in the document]
andreaangiolillo: if the signature changes we will have to update the resource code + the definition of the function here.

cfn-resources/teams/cmd/resource/team-user/team_user.go
```diff
@@ -0,0 +1,136 @@
...
+	// Iterate over the elements in A and add them to the toDelete array if they are not in B
+	for elem := range currentUsersSet {
+		if _, ok := newUsersSet[elem]; !ok {
+			toDelete = append(toDelete, elem)
+		}
+	}
+
+	// Return the two arrays
+	return toAdd, toDelete, nil
+}
```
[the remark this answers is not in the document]
marcosuma: sorry can you clarify? 

> what this function is testing

what do you mean with testing? the comment refers to the function implementation so I am not sure what you are referring to

cfn-resources/teams/cmd/resource/team-user/team_user.go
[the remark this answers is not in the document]
marcosuma: I don't see it as a concern tbh but thanks for raising the point.

Starting from the fact that changing the signature is a breaking change (as such it's already "a problem" on its own), I would say that once you edited the signature of the client, it doesn't take much effort to update the resource as well. To me the efforts are equal.

I would be more worried if _hypothetically_ this change would introduce a risk of missing an update which would cause a runtime error, which is not the case.

Said that, separating interfaces between components (i.e. making them less coupled) is a common practice. In theory for example if one day the client would change the `RemoveTeamUser` by requiring one more mandatory field you could always leave the `resource` code as it is and only change this interface by applying default values.

cfn-resources/teams/cmd/resource/team-user/team_user.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/mongodb/mongodbatlas-cloudformation-resources/util"
+	"github.com/mongodb/mongodbatlas-cloudformation-resources/util/logger"
+	atlasv2 "go.mongodb.org/atlas-sdk/v20231001001/admin"
+)
+
+type TeamUsersAPI interface {
+	GetUserByUsername(ctx context.Context, userName string) (*atlasv2.CloudAppUser, *http.Response, error)
+	AddTeamUser(ctx context.Context, orgID string, teamID string, addUserToTeam *[]atlasv2.AddUserToTeam) (*atlasv2.PaginatedApiAppUser, *http.Response, error)
+	RemoveTeamUser(ctx context.Context, orgID string, teamID string, userID string) (*http.Response, error)
+}
```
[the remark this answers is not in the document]
marcosuma: no it's not. This defines what the new interface is for (part of) the interaction between the `teams` resource and the `atlas client`.
If we want to take one extra step, we could actually add here all the methods that `teams` is using from the client and let this be the only interface you'd see in the `resource.go` code. I haven't gone that far yet, but we can do it.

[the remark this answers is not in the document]
andreaangiolillo: Thanks for the clarification

cfn-resources/teams/cmd/resource/team-user/team_user.go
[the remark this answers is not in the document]
andreaangiolillo: nvm. I misunderstood how the function was used

cfn-resources/teams/cmd/resource/team-user/team_user_test.go
```diff
@@ -0,0 +1,260 @@
...
+	mockAtlasV2Client.EXPECT().GetUserByUsername(context.Background(), validuser1).Return(&atlasv2.CloudAppUser{Id: &validuser1}, nil, nil)
+	mockAtlasV2Client.EXPECT().GetUserByUsername(context.Background(), invaliduser1).Return(nil, nil, errors.New("invalid username"))
+
+	// Call the FilterOnlyValidUsernames function
+	_, _, err := FilterOnlyValidUsernames(mockAtlasV2Client, usernames)
+
+	assert.NotNil(t, err)
+}
+
+func TestGetUserDeltas(t *testing.T) {
```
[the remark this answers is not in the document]
marcosuma: Name updated.

cfn-resources/teams/cmd/resource/team-user/team_user_test.go
```diff
@@ -0,0 +1,260 @@
...
+		if err != nil {
+			t.Errorf("Unexpected error: %v", err)
+		}
+
+		assert.Equal(t, testCase.expectedToAdd, toAdd)
+		assert.Equal(t, testCase.expectedToDelete, toDelete)
+	}
+}
+
+func TestUpdateTeamUsers(t *testing.T) {
```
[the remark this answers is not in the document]
marcosuma: I am following the `Test<MethodName>` name convention. Since the method we are testing `UpdateTeamUsers` then this test is called `TestUpdateTeamUsers` - yes, inside is using delete and create operation but it's an inner aspect of the method being tested.

cfn-resources/teams/cmd/resource/team-user/team_user.go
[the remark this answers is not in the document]
marcosuma: Resolving this one, if you have other questions we can re-open it.

cfn-resources/teams/cmd/resource/team-user/team_user_test.go
```diff
@@ -0,0 +1,261 @@
...
+		{Id: &user2},
+		{Id: &user3},
+	}
+
+	// Call the initUserSet function to create a map of user IDs
+	usersSet := initUserSet(users)
+
+	// Check that the map contains the expected user IDs
+	assert.True(t, usersSet[user1])
+	assert.True(t, usersSet[user2])
```
lantoli, line 46: this is the test assert style I'm more used to, and I'm not against it. But I think the more idiomatic Go for this is to use table style with an anonymous struct and a slice, e.g.: https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/mongodbatlas/util/type_conversion_test.go#L32..L69

cfn-resources/teams/cmd/resource/team-user/team_user.go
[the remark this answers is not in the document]
lantoli: I understand that the motivation to create this API is mostly to make mocking easier. In general I would agree that it's a good idea to have an interface that we own between the service and our code. 
But in this case it can be argued that we own the SDK so I kind of agree with Andrea concern that we'll have to maintain this interface in the middle when we're also really owners of the SDK and can influence on it. So I think that it doesn't add some much value as in other cases

lantoli (approved):
I like the general idea and also to improve tests by trying different things.
My only small concerns are related to specific details of implementation:
- if we should really create a new API interface when we already own the SDK
- test style seems more like other languages style than Go.

But I'm good to try this and see how it goes

cfn-resources/teams/cmd/resource/team-user/team_user.go
```diff
@@ -0,0 +1,136 @@
...
+
+func (s *TeamUsersAPIService) AddTeamUser(ctx context.Context, orgID string, teamID string, addUserToTeam *[]atlasv2.AddUserToTeam) (*atlasv2.PaginatedApiAppUser, *http.Response, error) {
+	return s.teamsAPI.AddTeamUser(ctx, orgID, teamID, addUserToTeam).Execute()
+}
+
+func (s *TeamUsersAPIService) RemoveTeamUser(ctx context.Context, orgID string, teamID string, userID string) (*http.Response, error) {
+	return s.teamsAPI.RemoveTeamUser(ctx, orgID, teamID, userID).Execute()
+}
+
+func FilterOnlyValidUsernames(mongoDBCloudUsersAPIClient TeamUsersAPI, usernames []string) ([]atlasv2.CloudAppUser, *http.Response, error) {
```
AgustinBettati, line 56: Since the intention of team_user and team_user_test is to define and test the business logic in `FilterOnlyValidUsernames`, `GetUsersToAddAndRemove`, and `UpdateTeamUsers` functions, the definition of `NewTeamUsersAPIService` with its default implementation could be defined in separate file `team_user_service`. The mock file could also be renamed to `team_user_service` to make more implicit what is being mocked.

AgustinBettati (approved):
Some feedback for possible improvements. My biggest open question is if we want to rely on GoMock moving forward vs having a dummy implementation on our side, but happy to move forward with this change and adjust later on if we have a strong opinion on the matter.

cfn-resources/teams/Makefile
```diff
@@ -17,3 +17,6 @@ debug:
 
 clean:
 	rm -rf bin
+
+unit:
```
AgustinBettati, line 21: are these tests being executed in CI? Would seem a slight adjustment in https://github.com/mongodb/mongodbatlas-cloudformation-resources/blob/master/Makefile#L44 is needed. (If change is not trivial we can handle in separate PR)

cfn-resources/teams/cmd/resource/team-user/team_user.go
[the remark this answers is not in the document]
marcosuma: > I understand that the motivation to create this API is mostly to make mocking easier

It starts because of that, yes. But in the end is not the only reason, but there is a [good answer](https://mongodb.stackenterprise.co/a/1065/747) by @gssbzn which I tend to agree with.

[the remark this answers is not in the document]
marcosuma: > when we're also really owners of the SDK and we can influence it

The relation between different packages and teams is an interesting one (I read an article when in Meta about our tendency of evolving teams based on how the product evolves, creating weird organizational structures - I don't remember it anymore which one it is).

My short answer is that we should never assume these org structures will remain the same forever. It is always possible to think that things will evolve in a way that a CFN integration and an SDK will be managed by not-so-close teams.

marcosuma:
> My biggest open question is if we want to rely on GoMock moving forward vs having a dummy implementation on our side, but happy to move forward with this change and adjust later on if we have a strong opinion on the matter.

@AgustinBettati the advantage of relying on something like GoMock is that you can set expectations that what was mocked has been called x times in a certain way. That's why probably is not the best (I wish I had the answer already) but definitely good to have for now

cfn-resources/test/mocks/team_user.go
```diff
@@ -0,0 +1,84 @@
+// Code generated by MockGen. DO NOT EDIT.
+// Source: cfn-resources/teams/cmd/resource/team-user/team_user.go
+
+// Package mock_teamuser is a generated GoMock package.
+package mock_teamuser
```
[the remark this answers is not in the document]
marcosuma: thank you!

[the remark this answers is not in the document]
marcosuma: comment addressed. Let me know if this works! @AgustinBettati

cfn-resources/test/mocks/team_user.go
```diff
@@ -0,0 +1,84 @@
+// Code generated by MockGen. DO NOT EDIT.
```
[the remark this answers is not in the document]
marcosuma: comment addressed. Let me know if this works! @AgustinBettati

cfn-resources/teams/Makefile
marcosuma: I actually realised that adding 

```
//go:build unit
```

on top of the `team_user_test.go`

I see this warning message on the `package teamuser`:

```
This file is within module "cfn-resources", which is not included in your workspace.
To fix this problem, you can add a go.work file that uses this directory.
See the documentation for more information on setting up your workspace:
https://github.com/golang/tools/blob/master/gopls/doc/workspace.md.
```

So I am going to remove this change and the makefile change and do it separately in another issue.

marcosuma: FYI @andreaangiolillo

cfn-resources/teams/cmd/resource/team-user/team_user_test.go
marcosuma: I agree with you but I think rewriting this in table style wouldn't change it much because the test case here is simple enough, it's just one case.  You can see below the others have been written in table style.

Resolving this one, feel free to re-open if you have a strong opinion.

cfn-resources/teams/cmd/resource/team-user/team_user.go
gssbzn, line 56: [nit] define an interface with only the method you need
```go
type UserGetter interface {
  GetUserByUsername
}
```

gssbzn, line 56:
```suggestion
func FilterOnlyValidUsernames(c TeamUsersAPI, usernames []string) ([]atlasv2.CloudAppUser, *http.Response, error) {
```

cfn-resources/teams/cmd/resource/team-user/team_user.go
```diff
@@ -0,0 +1,136 @@
...
+	return validUsers, nil, nil
+}
+
+func initUserSet(users []atlasv2.CloudAppUser) map[string]bool {
+	usersSet := make(map[string]bool)
+	for i := 0; i < len(users); i++ {
+		usersSet[*(users[i]).Id] = true
+	}
+	return usersSet
+}
```
gssbzn, line 45: prealloc
```suggestion
func initUserSet(users []atlasv2.CloudAppUser) map[string]bool {
	usersSet := make(map[string]interface{}, len(users))
	for _, u := range users {
		usersSet[u.GetId()] = true
	}
	return usersSet
}
```

cfn-resources/teams/cmd/resource/team-user/team_user.go
marcosuma: Done! I moved it to `team_user_service.go`. The other thing I did is to move the `mocks` folder close to the `team_user_service.go` file. So I just called it mocks. I am not sure if there are strong objections here though, but I think it could be a starting point. I felt that initiatlly I put it under `test` folder and it was a bit far from its real usage.

marcosuma: Resolving this conversation but please reopen if you don't agree with my choices.

cfn-resources/go.mod
```diff
@@ -36,6 +36,7 @@ require (
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/go-openapi/jsonpointer v0.19.6 // indirect
 	github.com/go-openapi/swag v0.22.4 // indirect
+	github.com/golang/mock v1.6.0
```
gssbzn, line 39: this is unused and probably a consequence of doing `go install` for having the bin
when needing only the bin consider doing
```
go install github.com/golang/mock/mockgen@latest
```
to avoid polluting your go.mod

gssbzn: I'm still going over this PR as I don't think you need mockgen yet

cfn-resources/teams/cmd/resource/team-user/mocks/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+## HOW-TO: Update mocks
+
+1. In order to update these mocks, you need to have `mockgen` **installed** on your machine.
+1. `cd` in the root of the `mongodbatlas-clouformation-resources` repo
+1. After that, you can run the `mockgen` command as following:
+
+```
+mockgen -source cfn-resources/teams/cmd/resource/team-user/team_user.go -destination /PATH/TO/REPOSITORY/mongodbatlas-cloudformation-resources/cfn-resources/teams/cmd/resource/team-user/mocks/mocks.go -package mocks
+```
```
gssbzn, line 13: [still considering if mockgen is really necessary] mockgen support go build tags which means you can and should use `go generate ` instead

cfn-resources/teams/cmd/resource/team-user/team_user.go
```diff
@@ -0,0 +1,106 @@
...
+
+func GetUsersToAddAndRemove(currentUsers []atlasv2.CloudAppUser, newUsers []atlasv2.CloudAppUser) (toAdd []string, toDelete []string, err error) {
```
gssbzn, line 47: [nit] names that include `And` or `Or` in their name do too many things, consider splitting into two methods that accept both sets

cfn-resources/teams/cmd/resource/team-user/team_user.go
```diff
@@ -0,0 +1,106 @@
...
+import (
+	"context"
+	"net/http"
+
+	"github.com/mongodb/mongodbatlas-cloudformation-resources/util"
+	"github.com/mongodb/mongodbatlas-cloudformation-resources/util/logger"
+	atlasv2 "go.mongodb.org/atlas-sdk/v20231001001/admin"
+)
+
+func FilterOnlyValidUsernames(mongoDBCloudUsersAPIClient TeamUsersAPI, usernames []string) ([]atlasv2.CloudAppUser, *http.Response, error) {
```
gssbzn, line 26: this is not filtering as much as it's validating as its currently used, it will fail at the first "not found user" and the whole function will error and the caller will exit with the error, can we either make sure this is the intended behavior and rename or fix the behavior to actually collect all valid names and operate over those?

cfn-resources/teams/cmd/resource/team-user/team_user.go
marcosuma: I think in team_user we need these three methods, so they are actually all needed. Are you suggesting one interface per method? I see that as a cons that will require more mocks and also more params passed to the methods.

cfn-resources/teams/cmd/resource/team-user/team_user.go
```diff
@@ -0,0 +1,106 @@
...
+}
+
+func UpdateTeamUsers(teamUsersAPIService TeamUsersAPI, existingTeamUsers *atlasv2.PaginatedApiAppUser, usernames []string, orgID, teamID string) error {
+	var newUsers []atlasv2.AddUserToTeam
+
+	validUsernames, _, err := FilterOnlyValidUsernames(teamUsersAPIService, usernames)
+	if err != nil {
+		return err
+	}
+	usersToAdd, usersToRemove, err := GetUsersToAddAndRemove(existingTeamUsers.Results, validUsernames)
```
gssbzn, line 81: listing is paginated and I see you are doing the call with the default of 100 per page, have you considered the case of a team having more than 100 users?

cfn-resources/teams/cmd/resource/team-user/team_user.go
```diff
@@ -0,0 +1,106 @@
...
+		if _, ok := newUsersSet[elem]; !ok {
+			toDelete = append(toDelete, elem)
+		}
+	}
+
+	// Return the two arrays
+	return toAdd, toDelete, nil
+}
+
+func UpdateTeamUsers(teamUsersAPIService TeamUsersAPI, existingTeamUsers *atlasv2.PaginatedApiAppUser, usernames []string, orgID, teamID string) error {
```
gssbzn, line 74: similar to my point on defining interfaces with exactly what you need this method should proabbly be split into the 3 operations it does, (1)validate, (2)remove, (3)add, which would let you have interfaces with only one method in them making them ideal and making your testing simpler

cfn-resources/teams/cmd/resource/team-user/team_user_test.go
```diff
@@ -0,0 +1,259 @@
...
+
+	// Check that the map does not contain any unexpected user IDs
+	assert.False(t, usersSet["user4"])
+	assert.False(t, usersSet["user5"])
+	assert.False(t, usersSet["user6"])
+}
+
+func TestFilterOnlyValidUsernames(t *testing.T) {
+	mockCtrl := gomock.NewController(t)
+	defer mockCtrl.Finish()
```
gssbzn, line 55: this is not necessary in the latest version of mackgen
```suggestion
```

cfn-resources/teams/cmd/resource/team-user/team_user_test.go
```diff
@@ -0,0 +1,259 @@
...
+
+	// Check that the returned slice contains only the valid usernames
+	assert.Equal(t, len(validUsers), 2)
+	assert.Equal(t, validuser1, *validUsers[0].Id)
+	assert.Equal(t, validuser2, *validUsers[1].Id)
+}
+
+func TestFilterOnlyValidUsernamesWithInvalidInput(t *testing.T) {
+	mockCtrl := gomock.NewController(t)
+	defer mockCtrl.Finish()
```
gssbzn, line 83:
```suggestion
```

cfn-resources/teams/cmd/resource/team-user/team_user_test.go
```diff
@@ -0,0 +1,259 @@
...
+
+	// Run test cases
+	for _, testCase := range testCases {
+		toAdd, toDelete, err := GetUsersToAddAndRemove(testCase.currentUsers, testCase.newUsers)
+		if err != nil {
+			t.Errorf("Unexpected error: %v", err)
+		}
+
+		assert.Equal(t, testCase.expectedToAdd, toAdd)
+		assert.Equal(t, testCase.expectedToDelete, toDelete)
```
gssbzn, line 154: please wrap your test case in t.Run
```suggestion
		t.Run(<test case name>, func(t *testing.T) {
			toAdd, toDelete, err := GetUsersToAddAndRemove(testCase.currentUsers, testCase.newUsers)
			require.NoError(err)
			assert.Equal(t, testCase.expectedToAdd, toAdd)
			assert.Equal(t, testCase.expectedToDelete, toDelete)
		}
```
if the test is thread safe consider also adding `t.Parallel()` inside the Run method

cfn-resources/teams/cmd/resource/team-user/team_user_test.go
```diff
@@ -0,0 +1,259 @@
...
+				return mockAtlasV2Client
+			},
+			existingTeamUsers: &atlasv2.PaginatedApiAppUser{Results: []atlasv2.CloudAppUser{{Id: &validuser2}}},
+			usernames:         []string{validuser1},
+			expectError:       false,
+		},
+	}
+
+	// Run test cases
+	for _, testCase := range testCases {
```
gssbzn, line 249: same as above, use `t.Run` for nested cases

cfn-resources/teams/cmd/resource/team-user/team_user.go
marcosuma: I am only changing the name here, agree it wasn't the best. I think it doesn't make sense to split it because it's actually logically connected (it has to calculate users to be removed and users to be added, it goes through the same lists and needs same initialization).

Resolving here, feel free to re-open if you disagree.

lantoli:
don't say we should classify in slow / no slow tests, just raising awareness about it: https://www.jvt.me/posts/2022/07/01/go-short-test/

cfn-resources/teams/cmd/resource/team-user/team_user.go
marcosuma: yes, I changed this on purpose because I believe this is the correct behaviour.
Now if the user inputs random text as `usernames` field, it will do best effort in finding usernames and then update the `teams` resource with whatever best effort users it found. This implies lots of "implicit unexpected" behaviour that customer may encounter if they actually accidentally made a typo in the username string..

so I agree we can change the name

cfn-resources/go.mod
marcosuma: may I ask you why is `unused`? I am using "github.com/golang/mock/gomock" in the `[...]_test.go` class

cfn-resources/teams/cmd/resource/team-user/team_user.go
marcosuma: That's a good call. While I would love to further improve this, I think this is another bug/behaviour from the original code. I am creating a follow-up task to update that accordingly.

https://jira.mongodb.org/browse/INTMDB-1280

Resolving, feel free to re-open if you have concerns.

cfn-resources/teams/cmd/resource/team-user/team_user.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+func (s *TeamUsersAPIService) RemoveTeamUser(ctx context.Context, orgID string, teamID string, userID string) (*http.Response, error) {
+	return s.teamsAPI.RemoveTeamUser(ctx, orgID, teamID, userID).Execute()
+}
+
+func FilterOnlyValidUsernames(mongoDBCloudUsersAPIClient TeamUsersAPI, usernames []string) ([]atlasv2.CloudAppUser, *http.Response, error) {
+	var validUsers []atlasv2.CloudAppUser
+	for _, elem := range usernames {
+		userToAdd, httpResp, err := mongoDBCloudUsersAPIClient.GetUserByUsername(context.Background(), elem)
```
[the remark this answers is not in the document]
marcosuma: that's another very good call thanks Gus! Again, if you look at the existing code, you see this was an existing (wrong) behaviour. The one of calling `Background()` every time seem to be very common in our code base.

I am guessing here that at the start of each of the CRUDL functions (e.g. `func Update(req handler.Request, prevModel *Model, currentModel *Model) (handler.ProgressEvent, error) {`) we should create a context and then pass that one to all the sub-methods call. 

**Am I right?** If so, I'd suggest here as well to create a further follow-up fix (even though if we are confident I can quickly fix this at least for the `Update` method).

https://jira.mongodb.org/browse/INTMDB-1281

cfn-resources/teams/cmd/resource/team-user/mocks/README.md
```diff
@@ -4,7 +4,7 @@ These mocks are generated through the `mockgen` command.
 
 ## HOW-TO: Update mocks
 
-1. In order to update these mocks, you need to have `mockgen` **installed** on your machine.
+1. In order to update these mocks, you need to have `mockgen` **installed** on your machine: `go install github.com/golang/mock/mockgen@latest`.
```
andreaangiolillo, line 7: Something to consider: https://github.com/golang/mock

> Update, June 2023: This repo and tool are no longer maintained. Please see [go.uber.org/mock](https://github.com/uber/mock) for a maintained fork instead.

We may want to update the make file `devtools` command to install mock gen as well.

cfn-resources/teams/cmd/resource/team-user/team_user_test.go
```diff
@@ -0,0 +1,267 @@
...
+		{Id: &user1},
+		{Id: &user2},
+		{Id: &user3},
+	}
+
+	// Call the initUserSet function to create a map of user IDs
+	usersSet := initUserSet(users)
+
+	// Check that the map contains the expected user IDs
+	assert.True(t, usersSet[user1].(bool))
```
gssbzn, line 44:
```suggestion
	assert.Contains(t, user1, usersSet)
```
some for the rest

cfn-resources/teams/cmd/resource/team-user/team_user_test.go
```diff
@@ -0,0 +1,267 @@
...
+	assert.True(t, usersSet[user2].(bool))
+	assert.True(t, usersSet[user3].(bool))
+
+	// Check that the map does not contain any unexpected user IDs
+	assert.Nil(t, usersSet["user4"])
```
gssbzn, line 49:
```suggestion
	assert.NotContains(t, usersSet, "user4")
```

cfn-resources/teams/cmd/resource/team-user/team_user_test.go
```diff
@@ -0,0 +1,267 @@
...
+	// Set up mock expectations for successful and unsuccessful calls to GetUserByUsername
+	mockAtlasV2Client.EXPECT().GetUserByUsername(context.Background(), validuser1).Return(&atlasv2.CloudAppUser{Id: &validuser1}, nil, nil)
+	mockAtlasV2Client.EXPECT().GetUserByUsername(context.Background(), validuser2).Return(&atlasv2.CloudAppUser{Id: &validuser2}, nil, nil)
+
+	// Call the ValidateUsernames function
+	validUsers, _, err := ValidateUsernames(mockAtlasV2Client, usernames)
+
+	if err != nil {
+		t.Errorf("Unexpected error: %v", err)
+	}
```
gssbzn, line 73:
```suggestion
	require.NoError(err)
```

cfn-resources/teams/cmd/resource/team-user/team_user_test.go
```diff
@@ -0,0 +1,267 @@
...
+
+	// Check that the returned slice contains only the valid usernames
+	assert.Equal(t, len(validUsers), 2)
```
gssbzn, line 76:
```suggestion
	assert.Len(t, validUsers, 2)
```

cfn-resources/teams/cmd/resource/team-user/team_user_test.go
```diff
@@ -0,0 +1,267 @@
...
+	assert.Equal(t, validuser1, *validUsers[0].Id)
```
gssbzn, line 77:
```suggestion
	assert.Equal(t, validuser1, validUsers[0].GetId())
```

cfn-resources/teams/cmd/resource/team-user/team_user_test.go
```diff
@@ -0,0 +1,267 @@
...
+	usernames := []string{validuser1, invaliduser1}
+
+	// Set up mock expectations for successful and unsuccessful calls to GetUserByUsername
+	mockAtlasV2Client.EXPECT().GetUserByUsername(context.Background(), validuser1).Return(&atlasv2.CloudAppUser{Id: &validuser1}, nil, nil)
+	mockAtlasV2Client.EXPECT().GetUserByUsername(context.Background(), invaliduser1).Return(nil, nil, errors.New("invalid username"))
+
+	// Call the ValidateUsernames function
+	_, _, err := ValidateUsernames(mockAtlasV2Client, usernames)
+
+	assert.NotNil(t, err)
```
gssbzn, line 98:
```suggestion
	require.NoError(t, err)
```

cfn-resources/teams/cmd/resource/team-user/team_user_test.go
```diff
@@ -0,0 +1,267 @@
...
+	validuser1 := "validuser1"
+	validuser2 := "validuser2"
+	invaliduser1 := "invaliduser1"
+
+	testCases := []struct {
+		testName             string
+		mockFuncExpectations func(*gomock.Controller) *mock.MockTeamUsersAPI
+		existingTeamUsers    *atlasv2.PaginatedApiAppUser
+		usernames            []string
+		expectError          bool
```
gssbzn, line 170:
```suggestion
		wantErr require.ErrorAssertionFunc
```
https://pkg.go.dev/github.com/stretchr/testify@v1.8.4/assert#ErrorAssertionFunc

cfn-resources/teams/cmd/resource/team-user/team_user_test.go
```diff
@@ -0,0 +1,267 @@
...
+
+	// Run test cases
+	for _, testCase := range testCases {
+		t.Run(testCase.testName, func(t *testing.T) {
+			mockCtrl := gomock.NewController(t)
+			client := testCase.mockFuncExpectations(mockCtrl)
+
+			err := UpdateTeamUsers(client, testCase.existingTeamUsers, testCase.usernames, "orgID", "teamID")
+
+			assert.Equal(t, testCase.expectError, err != nil)
```
gssbzn, line 264:
```suggestion
			tt.wantErr(t, UpdateTeamUsers(client, testCase.existingTeamUsers, testCase.usernames, "orgID", "teamID"))
```

cfn-resources/go.mod
gssbzn: oh you are right here, never mind me

cfn-resources/teams/Makefile
marcosuma: Resolving this one, please re-open if needed.

cfn-resources/teams/cmd/resource/team-user/team_user.go
marcosuma: I agree in part. Effectively the only (minor) improvement it could be done in this method is to move remove logic into its method

```
for ind := range usersToRemove {
		// remove user from team
		_, err := c.RemoveTeamUser(context.Background(), orgID, teamID, util.SafeString(&usersToRemove[ind]))
		if err != nil {
			return err
		}
	}
```

and add logic into its own
```
for ind := range usersToAdd {
		// add user to team
		newUsers = append(newUsers, atlasv2.AddUserToTeam{Id: util.SafeString(&usersToAdd[ind])})
	}
	// save all new users
	if len(newUsers) > 0 {
		_, _, err = c.AddTeamUser(context.Background(), orgID, teamID, &newUsers)
		if err != nil {
			return err
		}
	}
```

I say minor because the logic of these parts is quite easy. the `UpdateTeamUsers` works as a facade and it represents the business logic of `updating team users in the "teams" resource.

One might argue that `teamUsersAPIService` input param should be three different interfaces: one to get users, one to add and one to remove. I am not 100% sure of the final benefit of it. Maybe the only benefit is on the other methods that will receive their own dedicated interface, but I think the way I might have actually just solved that is by having converted those methods as private, which means they can only be called within this package (hence still be tested).

Ultimately only UpdateTeamUsers is the external interface that should be called and visible.

If you have a strong opinion I can separate them

cfn-resources/teams/cmd/resource/team-user/team_user_test.go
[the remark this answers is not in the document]
marcosuma: resolving this comment, feel free to open if it's still not clear.

cfn-resources/teams/cmd/resource/team-user/mocks/README.md
marcosuma: Thank you Andrea, I was aware of (left a comment at the start of this PR) and first wanted to check if there were better ideas. I updated to `go.uber.org/mock` now.

> We may want to update the make file devtools command to install mock gen as well.

I am not sure what you mean here but I might have followed a better approach with `go:generate` as suggested by @gssbzn below, as such this readme will be removed.

cfn-resources/teams/cmd/resource/team-user/mocks/README.md
marcosuma: I am using go:generate now

Makefile
```diff
@@ -47,3 +47,7 @@ unit-test:
 .PHONY: update-atlas-sdk
 update-atlas-sdk: ## Update the atlas-sdk dependency
 	(cd cfn-resources && ./scripts/update-sdk.sh)
+
+.PHONY: generate-mocks
```
marcosuma, line 51: @gssbzn added go:generate to the file and also this in the Makefile. I am not sure if we want to further automate this? (e.g. running it on every build?)

cfn-resources/teams/cmd/resource/team-user/team_user.go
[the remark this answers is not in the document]
marcosuma: resolving this conversation, it won't be lost since I am creating follow-up issues. Feel free to re-open if you want to further discuss.

cfn-resources/teams/cmd/resource/team-user/team_user.go
marcosuma: I moved these methods as private and provided an explanation below https://github.com/mongodb/mongodbatlas-cloudformation-resources/pull/810#discussion_r1391093913 - resolving this, we can continue in the other comment.

Makefile
gssbzn: I have no strong opinion here but from experience I found mocks to get out of sync or even have bugs because people would forget to regenerate an interface so we ended up having a test to fail the CI if there were diffs when running this target
https://github.com/mongodb/mongodb-atlas-cli/blob/cabf9a65a6bc7f6602d96f58689e9f21005e4941/.github/workflows/code-health.yml#L114-L137

Also to keep in mind the generation doesn't delete unused files but those are less distracting that having the generate being on a broken state and only relaizing when you need it

cfn-resources/teams/cmd/resource/team-user/team_user.go
```diff
@@ -0,0 +1,106 @@
...
+	validUsernames, _, err := validateUsernames(c, usernames)
+	if err != nil {
+		return err
+	}
+	usersToAdd, usersToRemove, err := getChangesForTeamUsers(existingTeamUsers.Results, validUsernames)
+	if err != nil {
+		return err
+	}
+
+	for ind := range usersToRemove {
```
gssbzn, line 38: [q] `ind` for index? but wouldn't this be `key` since a map?

gssbzn (approved):
LGTM thanks for all the work here 💯

cfn-resources/teams/cmd/resource/team-user/mocks/README.md
marcosuma: ok added to the `make tools` - thank you!

AgustinBettati (approved):
LGTM, thanks for addressing comments. Great improvement here.

PR #828, opened 2023-11-24
chore: fix DatabaseUser examples 

oarbusi:
## Proposed changes

Fix examples to avoid issues when creating DatabaseUser stack in Cloudformation using the example. The current examples failed with the error:
`Requested attribute Username must be a readonly property in schema for MongoDB::Atlas::DatabaseUser. Rollback requested by user.` 

## Type of change:

  expected)

## Manual QA performed:

- [X] cfn invoke for each of CRUDL/cfn test
- [X] Updated resource in  [example](https://github.com/mongodb/mongodbatlas-cloudformation-resources/tree/master/examples)
- [X] Published to AWS private registry
- [ ] Validated in Atlas UI
- [ ] Included screenshots

## Required Checklist:

  works in Atlas


examples/database-user/ldapUser.json
```diff
@@ -83,15 +83,5 @@
...
       }
     }
-  },
-  "Outputs": {
-    "MongoDBIamUsername": {
-      "Value": {
-        "Fn::GetAtt": [
-          "ldapUser",
-          "Username"
-        ]
```
andreaangiolillo, line 93: the reason why we added this section was to show customers how they could output information about the cfn resource. I don't think we should remove it

cfn-resources/database-user/mongodb-atlas-databaseuser.json
```diff
@@ -140,7 +142,8 @@
       },
       "minItems": 1,
       "type": "array",
-      "uniqueItems": true
+      "uniqueItems": true,
+      "insertionOrder": false
```
andreaangiolillo, line 146: why is insertionOrder needed? doesn't it default to false?

andreaangiolillo (commented):
I think we could reach out to aws to understand if something has changed with how primary identifiers work and see what is the recommended way to print them with the outputs

cfn-resources/database-user/mongodb-atlas-databaseuser.json
oarbusi: it defaults to true as you can see in the [docs](https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-schema.html)

examples/database-user/ldapUser.json
oarbusi: I understand the reasoning, but as of now, the examples, when creating the resources, fail with the error specified in the PR description. I think the error can lead to a lot of confusion for customers.

lantoli: I think for sure it's better to have a working example. In this case I understand that it's interesting to show something in the Outputs, but Username is a param provided by the user so they already know its value. And in this resource I think they provide all values so there anything interesting to be returned by the resource which might be useful in the output, I think

cfn-resources/database-user/mongodb-atlas-databaseuser.json
oarbusi: now that I think about it, the behaviour is changed with this change, because default is true and now I'm setting it to false even though I think it makes sense to be false. I can change it to set it to the default value.

examples/database-user/ldapUser.json
lantoli: I think a great example to keep a param in ouputs is in iam-aws example, UserCFNIdentifier that is calculated in the resource.

cfn-resources/database-user/mongodb-atlas-databaseuser.json
```diff
@@ -131,7 +132,8 @@
...
```
lantoli, line 136: not strongly against this but I'd probably just remove the cfn generate warnings at the moment, using the default value so we don't change the current behavior

lantoli: as this is the only change that requires a release, I'd probably revert this change so we only change README and examples and we don't need a release

lantoli (approved):

lantoli (approved):

PR #892, opened 2024-01-31
chore: New released Atlas Go SDK can't be used until several hours later

lantoli:
## Description

CLOUDP-221977

Using latest is known to cause cache issues with go.dev, using directly the tag should help reduce those problems.

same change as: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/1885


gssbzn (approved):
LGTM

PR #1012, opened 2024-04-24
fix: Allows not setting optional attributes DataProcessRegion and CloudProviderConfig for `FederatedDatabaseInstance`

oarbusi:
## Proposed changes

Fixes bug where creation of resource was not possible without setting values for optional attributes DataProcessRegion and CloudProviderConfig.
Now it's possible to create a `FederatedDatabaseInstance` without setting those attributes

Some opportunistic improvements in the code have been maded in the resource.

Link to any related issue(s): CLOUDP-219922

## Type of change:

  expected)

## Manual QA performed:

- [x] Validated in Atlas UI
- [x] Included screenshots

## Required Checklist:

  works in Atlas

## Further comments
Tests results:
Screenshot 2024-04-24 at 09 42 38
Stack creation and update:
Screenshot 2024-04-24 at 09 55 29
Atlas UI after creation:
Screenshot 2024-04-24 at 09 55 38

### DIFF
--- a/cfn-resources/federated-database-instance/cmd/resource/resource.go
+++ b/cfn-resources/federated-database-instance/cmd/resource/resource.go
@@ -28,13 +28,12 @@ import (
 	"github.com/mongodb/mongodbatlas-cloudformation-resources/util/logger"
 	progress_events "github.com/mongodb/mongodbatlas-cloudformation-resources/util/progressevent"
 	"github.com/mongodb/mongodbatlas-cloudformation-resources/util/validator"
-	"github.com/spf13/cast"
-	atlasSDK "go.mongodb.org/atlas-sdk/v20231115002/admin"
+	atlasSDK "go.mongodb.org/atlas-sdk/v20231115008/admin"
 )
 
-var CreateRequiredFields = []string{constants.ProjectID, constants.TenantName, constants.DataFederationRoleID, constants.DataFederationTestS3Bucket, constants.DataProcessRegion}
+var CreateRequiredFields = []string{constants.ProjectID, constants.TenantName}
 var ReadRequiredFields = []string{constants.ProjectID, constants.TenantName}
-var UpdateRequiredFields = []string{constants.ProjectID, constants.TenantName, constants.DataFederationRoleID, constants.DataFederationTestS3Bucket, constants.SkipRoleValidation}
+var UpdateRequiredFields = []string{constants.ProjectID, constants.TenantName}
 var DeleteRequiredFields = []string{constants.ProjectID, constants.TenantName}
 var ListRequiredFields = []string{constants.ProjectID}
 
@@ -69,8 +68,12 @@ func Create(req handler.Request, prevModel *Model, currentModel *Model) (handler
 	if peErr != nil {
 		return *peErr, nil
 	}
-	createFederatedDatabaseAPIRequest := client.Atlas20231115002.DataFederationApi.CreateFederatedDatabase(context.Background(), *currentModel.ProjectId, &dataLakeTenantInput)
-	dataLakeTenant, response, err := createFederatedDatabaseAPIRequest.Execute()
+	dataLakeTenant, response, err := client.AtlasSDK.DataFederationApi.CreateFederatedDatabaseWithParams(
+		context.Background(),
+		&atlasSDK.CreateFederatedDatabaseApiParams{
+			GroupId:        *currentModel.ProjectId,
+			DataLakeTenant: &dataLakeTenantInput,
+		}).Execute()
 
 	defer closeResponse(response)
 	if err != nil {
@@ -78,7 +81,6 @@ func Create(req handler.Request, prevModel *Model, currentModel *Model) (handler
 	}
 	readModel := Model{ProjectId: currentModel.ProjectId, TenantName: currentModel.TenantName, Profile: currentModel.Profile}
 	readModel.getDataLakeTenant(*dataLakeTenant)
-
 	return handler.ProgressEvent{
 		OperationStatus: handler.Success,
 		Message:         "Create Completed",
@@ -104,8 +106,7 @@ func Read(req handler.Request, prevModel *Model, currentModel *Model) (handler.P
 		return *peErr, nil
 	}
 
-	getFederatedDatabaseAPIRequest := client.Atlas20231115002.DataFederationApi.GetFederatedDatabase(context.Background(), *currentModel.ProjectId, *currentModel.TenantName)
-	dataLakeTenant, response, err := getFederatedDatabaseAPIRequest.Execute()
+	dataLakeTenant, response, err := client.AtlasSDK.DataFederationApi.GetFederatedDatabase(context.Background(), *currentModel.ProjectId, *currentModel.TenantName).Execute()
 
 	defer closeResponse(response)
 	if err != nil {
@@ -139,7 +140,12 @@ func Update(req handler.Request, prevModel *Model, currentModel *Model) (handler
 		return *peErr, nil
 	}
 
-	updateFederatedDatabaseAPIRequest := client.Atlas20231115002.DataFederationApi.UpdateFederatedDatabase(context.Background(), *currentModel.ProjectId, *currentModel.TenantName, &dataLakeTenantInput)
+	_, checkExistsResponse, checkExistsErr := client.AtlasSDK.DataFederationApi.GetFederatedDatabase(context.Background(), *currentModel.ProjectId, *currentModel.TenantName).Execute()
+	if checkExistsErr != nil {
+		return handleError(checkExistsResponse, UPDATE, checkExistsErr)
+	}
+
+	updateFederatedDatabaseAPIRequest := client.AtlasSDK.DataFederationApi.UpdateFederatedDatabase(context.Background(), *currentModel.ProjectId, *currentModel.TenantName, &dataLakeTenantInput)
 	updateFederatedDatabaseAPIRequest = updateFederatedDatabaseAPIRequest.SkipRoleValidation(*currentModel.SkipRoleValidation)
 	dataLakeTenant, response, err := updateFederatedDatabaseAPIRequest.Execute()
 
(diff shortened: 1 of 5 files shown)


lantoli:
contract test is failing

cfn-resources/federated-database-instance/cmd/resource/resource.go
```diff
@@ -28,13 +28,12 @@ import (
 	"github.com/mongodb/mongodbatlas-cloudformation-resources/util/logger"
 	progress_events "github.com/mongodb/mongodbatlas-cloudformation-resources/util/progressevent"
 	"github.com/mongodb/mongodbatlas-cloudformation-resources/util/validator"
-	"github.com/spf13/cast"
-	atlasSDK "go.mongodb.org/atlas-sdk/v20231115002/admin"
+	atlasSDK "go.mongodb.org/atlas-sdk/v20231115008/admin"
```
lantoli, line 31: if using latest SDK please remove package alias atlasSDK so you can use just admin. CreateFederatedDatabaseApiParams, etc.

cfn-resources/federated-database-instance/cmd/resource/resource.go
```diff
@@ -28,13 +28,12 @@ import (
...
 )
 
-var CreateRequiredFields = []string{constants.ProjectID, constants.TenantName, constants.DataFederationRoleID, constants.DataFederationTestS3Bucket, constants.DataProcessRegion}
+var CreateRequiredFields = []string{constants.ProjectID, constants.TenantName}
 var ReadRequiredFields = []string{constants.ProjectID, constants.TenantName}
-var UpdateRequiredFields = []string{constants.ProjectID, constants.TenantName, constants.DataFederationRoleID, constants.DataFederationTestS3Bucket, constants.SkipRoleValidation}
+var UpdateRequiredFields = []string{constants.ProjectID, constants.TenantName}
```
lantoli, line 36: you can have just one var CreateReadUpdateDeleteRequiredFields, or just call it RequiredFields, apart from ListRequiredFields

cfn-resources/federated-database-instance/cmd/resource/resource.go
```diff
@@ -69,16 +68,19 @@ func Create(req handler.Request, prevModel *Model, currentModel *Model) (handler
...
-	createFederatedDatabaseAPIRequest := client.Atlas20231115002.DataFederationApi.CreateFederatedDatabase(context.Background(), *currentModel.ProjectId, &dataLakeTenantInput)
-	dataLakeTenant, response, err := createFederatedDatabaseAPIRequest.Execute()
+	dataLakeTenant, response, err := client.AtlasSDK.DataFederationApi.CreateFederatedDatabaseWithParams(
+		context.Background(),
+		&atlasSDK.CreateFederatedDatabaseApiParams{
+			GroupId:        *currentModel.ProjectId,
+			DataLakeTenant: &dataLakeTenantInput,
+		}).Execute()
 
 	defer closeResponse(response)
```
lantoli, line 78: uhm, this defer seems weird, do we need it?

cfn-resources/federated-database-instance/cmd/resource/resource.go
```diff
@@ -241,31 +244,46 @@ func handleError(response *http.Response, method string, err error) (handler.Pro
 			Message:          fmt.Sprintf("%s:%s", method, err.Error()),
 			HandlerErrorCode: cloudformation.HandlerErrorCodeAlreadyExists}, nil
 	}
+	if response.StatusCode == http.StatusNotFound {
```
lantoli, line 247: is this a new check that was needed before?

lantoli (dismissed):

cfn-resources/federated-database-instance/cmd/resource/resource.go
oarbusi: Done!

cfn-resources/federated-database-instance/cmd/resource/resource.go
oarbusi: I tried to change as little as possible the existing code. Looking at the http.Response entity, you can see:
_"The http Client and Transport guarantee that Body is always non-nil, even on responses without a body or responses with a zero-length body. It is the caller's responsibility to close Body. The default HTTP client's Transport may not reuse HTTP/1.x "keep-alive" TCP connections if the Body is not read to completion and closed."_

cfn-resources/federated-database-instance/cmd/resource/resource.go
oarbusi: yes, update operation was not checking if resource to be updated existed before updating, so I added that check before updating and the handling of the 404 status here

cfn-resources/federated-database-instance/cmd/resource/resource.go
```diff
@@ -241,42 +244,60 @@ func handleError(response *http.Response, method string, err error) (handler.Pro
...
-	if model.DataProcessRegion.CloudProvider != nil {
-		cloudProvider = *model.DataProcessRegion.CloudProvider
-	}
-	dataLakeTenant = atlasSDK.DataLakeTenant{
-		Name: model.TenantName,
-		CloudProviderConfig: &atlasSDK.DataLakeCloudProviderConfig{
-			Aws: atlasSDK.DataLakeAWSCloudProviderConfig{
-				TestS3Bucket: *model.CloudProviderConfig.TestS3Bucket,
-				RoleId:       *model.CloudProviderConfig.RoleId,
+func (model *Model) setDataLakeTenant() (dataLakeTenant admin.DataLakeTenant) {
```
EspenAlbert, line 258: Unit tests candidate?

cfn-resources/federated-database-instance/cmd/resource/resource.go
oarbusi: I thought about simplifying this but I prefer to keep it as it is for:
- Consistency with other resources
- Possible future evolution of required fields
- You can see very quickly which are the required fields for each operation, even if they are the same
- Not a fan of `CreateReadUpdateDeleteRequiredFields ` 😅

cfn-resources/federated-database-instance/cmd/resource/resource.go
```diff
@@ -241,42 +244,60 @@ func handleError(response *http.Response, method string, err error) (handler.Pro
...
+func (model *Model) newDataFederationDataStorage() *admin.DataLakeStorage {
+	return &admin.DataLakeStorage{
 		Databases: model.newDataFederationDatabase(),
 		Stores:    model.newStores(),
 	}
 }
 
-func (model *Model) newDataFederationDatabase() []atlasSDK.DataLakeDatabaseInstance {
+func (model *Model) newDataFederationDatabase() *[]admin.DataLakeDatabaseInstance {
 	if model.Storage == nil {
```
EspenAlbert, line 301: Maybe return empty list instead of nil?  (`[]admin.DataLakeDatabaseInstance{}`)
Related to the SDK upgrade where `nil` pointers will not be dumped in the json. Leading to no updates when Databases might be set to nil? However, might not apply to this resource

oarbusi: In this case this is used in create and update operations. I think that if they are not set in the configuration, we want them not to be in the json so that they are not set in the create or update

cfn-resources/federated-database-instance/cmd/resource/resource.go
```diff
@@ -241,42 +244,60 @@ func handleError(response *http.Response, method string, err error) (handler.Pro
...
+	if dataProcessRegion != nil && dataProcessRegion.CloudProvider != nil && dataProcessRegion.Region != nil {
+		dataLakeTenant.DataProcessRegion.CloudProvider = *dataProcessRegion.CloudProvider
+		dataLakeTenant.DataProcessRegion = &admin.DataLakeDataProcessRegion{
+			CloudProvider: *dataProcessRegion.CloudProvider,
+			Region:        *dataProcessRegion.Region,
+		}
 	}
+
+	if dataProcessRegion != nil && dataProcessRegion.Region != nil {
+		dataLakeTenant.DataProcessRegion = &admin.DataLakeDataProcessRegion{
```
AgustinBettati, line 284: Can we have the scenario where a users defines `DataProcessRegion` with only the `CloudProvider` defined? Would we want to still define data process region in this case or was this causing issues?

oarbusi: For the API, when `DataProcessRegion` is defined both `Region` and `CloudProvider` are required. In the case of CFN the default is AWS so that means that in this case `Region` can be specified without specifying `CloudProvider`. This if is handling this specific edge case of CFN

oarbusi: > where `CloudProvider` is defined but not `Region`
As per the API this should not be possible, both are required. I see that the current implementation would ignore `DataProcessRegion` all together in that case. I would not put it as required. 
Question is how we want to handle this case: ignore `DataProcessRegion`(as it is now) or throw an error? Note that his method is called from both Create and Update operations.

cfn-resources/federated-database-instance/cmd/resource/resource.go
lantoli: i understand but also we have to be good boyscouts :-)

we never close http responses for SDK API calls, I'd like to understand why it's different here

cfn-resources/federated-database-instance/cmd/resource/resource.go
AgustinBettati: I would think from the users point of view, what documentation or error messages are in place to understand that if `DataProcessRegion` is defined, then `Region` must be defined and `CloudProvider` is optional as it has the default value? With this in mind we can adjust if needed.

oarbusi: I would say that these 2 docs ( [first](https://github.com/mongodb/mongodbatlas-cloudformation-resources/blob/master/cfn-resources/federated-database-instance/docs/README.md#dataprocessregion), [second](https://github.com/mongodb/mongodbatlas-cloudformation-resources/blob/master/cfn-resources/federated-database-instance/docs/dataprocessregion.md)) should be enough. Also it explicitly says that only supported cloud provider is AWS. WDYT?

cfn-resources/federated-database-instance/cmd/resource/resource.go
oarbusi: Went deeper into it. Basically all clients have to close the body if you want to [avoid resource leaks](https://stackoverflow.com/questions/33238518/what-could-happen-if-i-dont-close-response-body). The good thing is that the autogenerated SDK closes the body after request ha been done. I removed it from this resource.

cfn-resources/federated-database-instance/cmd/resource/resource.go
oarbusi: Adressed making Region required when DataProcessRegion is defined

AgustinBettati (approved):
LGTM

PR #1052, opened 2024-05-28
fix: Fully supports cidr block property in access-list-api-key resource and upgrade to new sdk version

AgustinBettati:
## Proposed changes

_Jira ticket:_ CLOUDP-250920 (fix of CIDR block), and part of CLOUDP-222755 for updating sdk version.

When upgrading SDK it was clear CIDR block property was not being supported properly. Fixed implementation using our terraform implementation as guidance, add a contract test for the property (verified this test was failing before the fix).

## Type of change:

  expected)

## Manual QA performed:

- [ ] Validated in Atlas UI
- [ ] Included screenshots

## Required Checklist:

  works in Atlas


cfn-resources/access-list-api-key/cmd/resource/resource.go
```diff
@@ -227,6 +214,23 @@ func Delete(req handler.Request, prevModel *Model, currentModel *Model) (handler
 		ResourceModel:   nil}, nil
 }
 
+// If CIDR block is defined, subnet mask is removed so get and delete requests work correctly
```
AgustinBettati, line 217: fix inspired from https://github.com/mongodb/terraform-provider-mongodbatlas/blob/master/internal/service/accesslistapikey/resource_access_list_api_key.go#L84. Also reached out to IAM team to confirm the current behavior (or limitation) of [getApiKeyAccessList](https://www.mongodb.com/docs/atlas/reference/api-resources-spec/v2/#tag/Programmatic-API-Keys/operation/getApiKeyAccessList) which cannot receive the cidr block value directly.

lantoli: I think this is not the same implementation.  In TF code, the switch only enters one path, so it's like using an if else, not 2 continuous if statements.

also note that the TF code could be quite simplified to: 
```

if  CIDRBlock != "" {
		parts := strings.SplitN(CIDRBlock, "/", 2)
		if parts[1] == "32" {
			entry = parts[0]
		} else {
			entry = CIDRBlock
		}
} else {
		entry = IPAddress
}
```
i think we could use that idea here as well

lantoli: can CidrBlock and IPAddress exist at the same time?

cfn-resources/access-list-api-key/cmd/resource/resource.go
```diff
@@ -89,8 +90,8 @@ func Create(req handler.Request, prevModel *Model, currentModel *Model) (handler
 	}
 
 	// createReq.ApiService.
-	entryList := make([]atlasSDK.UserAccessList, 0)
-	var access atlasSDK.UserAccessList
+	entryList := make([]admin.UserAccessListRequest, 0)
+	var access admin.UserAccessListRequest
 	if currentModel.CidrBlock != nil {
```
lantoli, line 95: can CidrBlock and IPAddress exist at the same time?

lantoli (approved):

cfn-resources/access-list-api-key/cmd/resource/resource.go
AgustinBettati: They cannot, it is being validated in the create [here](https://github.com/mongodb/mongodbatlas-cloudformation-resources/blob/master/cfn-resources/access-list-api-key/cmd/resource/resource.go#L84). One must be defined, and both cannot be defined at the same time. Having an else should be the same as checking that ip address is defined.

cfn-resources/access-list-api-key/cmd/resource/resource.go
AgustinBettati: we are throwing an error [here](https://github.com/mongodb/mongodbatlas-cloudformation-resources/blob/master/cfn-resources/access-list-api-key/cmd/resource/resource.go#L84) in the create (and update is not supported in this resource).

cfn-resources/access-list-api-key/cmd/resource/resource.go
lantoli: nice, thx

PR #1079, opened 2024-06-18
feat: Adds GlobalClusterSelfManagedSharding to Cluster

EspenAlbert:
## Proposed changes

Adds GlobalClusterSelfManagedSharding to Cluster and a new example: `cluster-self-managed-sharding.json`

Link to any related issue(s): CLOUDP-256267

## Type of change:

  expected)

## Manual QA performed:

- [x] Validated in Atlas UI
- [x] Included screenshots

## Required Checklist:

  works in Atlas

## Further comments

Screenshots:

Atlas UI of example:

image

contract tests:

image

CFN stack (see [Jira Ticket](https://jira.mongodb.org/browse/CLOUDP-256534) for fixing updates to replication spec):

image


cfn-resources/cluster/mongodb-atlas-cluster.json
```diff
@@ -341,6 +341,10 @@
         "NONE"
       ]
     },
+    "GlobalClusterSelfManagedSharding": {
+      "description": "Set this field to configure the Sharding Management Mode when creating a new Global Cluster. When set to false, the management mode is set to Atlas-Managed Sharding. This mode fully manages the sharding of your Global Cluster and is built to provide a seamless deployment experience. When set to true, the management mode is set to Self-Managed Sharding. This mode leaves the management of shards in your hands and is built to provide an advanced and flexible deployment experience. This setting cannot be changed once the cluster is deployed",
```
lantoli, line 345: description is fine but I'm working in adding a little bit more of detail about how to configure, it should be fast, i'll keep you posted.
Please wait for merge, or alternatively you can merge and do the doc changes in another PR

cfn-resources/cluster/cmd/resource/mappings.go
```diff
@@ -39,6 +39,7 @@ func mapClusterToModel(model *Model, cluster *admin.AdvancedClusterDescription)
 	model.CreatedDate = util.TimePtrToStringPtr(cluster.CreateDate)
 	model.DiskSizeGB = cluster.DiskSizeGB
 	model.EncryptionAtRestProvider = cluster.EncryptionAtRestProvider
+	model.GlobalClusterSelfManagedSharding = cluster.GlobalClusterSelfManagedSharding
```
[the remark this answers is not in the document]
lantoli: i have mixed feelings here. On the one hand I think it's a good idea to have contract tests about as many features and resources as possible.
on the other hand this feature requires GEOSHARDED clusters that are very slow to create, update, etc., and we'll be probably testing more Atlas than the CFN logic, so in this case I'm not sure about the pros/cons balance

lantoli (approved):

cfn-resources/cluster/mongodb-atlas-cluster.json
lantoli: @EspenAlbert I have a proposal doc in this PR: https://github.com/mongodb/terraform-provider-mongodbatlas/pull/2348
WDYT about waiting until doc team reviews?

EspenAlbert: Sounds good 👍

cfn-resources/cluster/cmd/resource/mappings.go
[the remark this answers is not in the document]
maastha: I understand but I'm not sure if that's a good enough reason to not have a test at all. Looking at the current [inputs](https://github.com/mongodb/mongodbatlas-cloudformation-resources/blob/master/cfn-resources/cluster/test/inputs_1_create.template.json) file for cluster CFN resource, it looks like there is very minimal coverage as it is and only one REPLICASET test case with only one replication spec is being tested.

Also I don't see cluster tests running in CI.

I'd suggest we should have a test to avoid further divergence from the resource. Since the test doesn't run in CI anyway it would be good to at least have the option to run it locally.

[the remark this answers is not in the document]
lantoli: if there are not e2e or contract tests cluster tests we should definitely have.

my point was that we should have only some basic tests in e2e / contract tests, not all different combinations of parameters and features, that can normally better tested in unit tests.

in this case I'm not sure about having a specific e2e or contract test only to test this attribute, but I agree to have some e2e and contract tests running in CI for cluster.

cfn-resources/cluster/mongodb-atlas-cluster.json
lantoli: doc is already reviewed in that PR, you can already take the doc from there

EspenAlbert: Nice. Updated in [781c02c](https://github.com/mongodb/mongodbatlas-cloudformation-resources/pull/1079/commits/781c02ce2c66d3646210a28804c9791a539a0b37)
Need a restamp

cfn-resources/cluster/cmd/resource/mappings.go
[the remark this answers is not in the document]
EspenAlbert: Added a follow-up ticket [here](https://jira.mongodb.org/browse/CLOUDP-256730)

lantoli (dismissed):

cfn-resources/cluster/mongodb-atlas-cluster.json
```diff
@@ -341,6 +341,10 @@
         "NONE"
       ]
     },
+    "GlobalClusterSelfManagedSharding": {
+      "description": "Flag that indicates if cluster uses Atlas-Managed Sharding (false) or Self-Managed Sharding (true).",
```
lantoli, line 345: as this is the only doc, I'll probably go for the long explanation (the one in the TF resource)

cfn-resources/cluster/docs/README.md
```diff
@@ -139,6 +141,16 @@ _Allowed Values_: <code>AWS</code> | <code>GCP</code> | <code>AZURE</code> | <co
 
 _Update requires_: [No interruption](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-update-behaviors.html#update-no-interrupt)
 
+#### GlobalClusterSelfManagedSharding
+
+Flag that indicates if cluster uses Atlas-Managed Sharding (false) or Self-Managed Sharding (true).
```
lantoli, line 146: same, as this is the only doc, I'll probably go for the long explanation (the one in the TF resource)

cfn-resources/cluster/mongodb-atlas-cluster.json
```diff
@@ -446,7 +450,8 @@
   "createOnlyProperties": [
     "/properties/Name",
     "/properties/ProjectId",
-    "/properties/Profile"
+    "/properties/Profile",
+    "/properties/GlobalClusterSelfManagedSharding"
```
AgustinBettati, line 454: q: Does defining this property as createOnly have any impact when `GlobalClusterSelfManagedSharding` is left empty and then populated with value coming from API?

cfn-resources/cluster/docs/README.md
```diff
@@ -139,6 +141,16 @@ _Allowed Values_: <code>AWS</code> | <code>GCP</code> | <code>AZURE</code> | <co
...
+
+_Required_: No
+
+_Type_: Boolean
+
+_Update requires_: [Replacement](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-update-behaviors.html#update-replacement)
```
AgustinBettati, line 152: similar to the concern mentioned in terraform, it would seem that marking property as `createOnly` forces a replacement if modified. Is this intentional? We could consider leaving as a regular property. cc: @lantoli

cfn-resources/cluster/mongodb-atlas-cluster.json
EspenAlbert: The CFN implementation differs a bit from Terraform:
1. If the value is `nil` then we use the value from the API
2. else, we use the value specified in the model

The normal test case/examples don't specify this flag and still managed the full CRUD operations, so I believe we should be safe 👍

cfn-resources/cluster/docs/README.md
EspenAlbert: See [comment above](https://github.com/mongodb/mongodbatlas-cloudformation-resources/pull/1079#discussion_r1645675545)

lantoli: @EspenAlbert I think it's useful to do a manual test of creating a stack and then modifying the param and update the stack to see if we get an error (correct) instead of a cluster replacement (bad)

cfn-resources/cluster/cmd/resource/mappings.go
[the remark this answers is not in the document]
maastha: >not all different combinations of parameters and features, that can be normally better tested in unit tests.

I'm not suggesting to have tests for all combinations of parameters, but at least for different types of cluster, I would consider that basic coverage. 
In this case, the type would be GEOSHARDED using this attribute.

cfn-resources/cluster/docs/README.md
EspenAlbert: image

I switched it from `null` to `false` and got the error message above (no cluster deletion) ☝️ 

Then I did a different update (setting `PitEnabled: false` (was true)) and it worked (see screenshot 👇 )

image

Can you confirm this is the correct behavior @lantoli ?

marcosuma: Good catch, I wonder if that works as expected or it's an unexpected API behavior? 

Also @EspenAlbert any chance to convert this into a test instead of leaving it manual?

EspenAlbert: @marcosuma 
To me it is expected behavior ✅  but I want to check with @lantoli before merging.
Yes, plan on adding that in a separate ticket, see [discussion above](https://github.com/mongodb/mongodbatlas-cloudformation-resources/pull/1079#discussion_r1645469457)

lantoli: @EspenAlbert thanks, can you do the last test of starting with attribute true, and then try to change to false and if it fails or the cluster is recreated?

lantoli: @EspenAlbert just to be completely sure (in case null is treated differently), can you do 2 more tests, one creating with false and trying to change to true, and the opposite, starting with true and change to false? thanks!

lantoli (approved):

cfn-resources/cluster/docs/README.md
EspenAlbert: Done ✅ 
Getting the same error in both cases

image

PR #1087, opened 2024-06-26
fix: Edits to replication specs for cluster

EspenAlbert:
## Proposed changes

1. Adds e2e test for cluster
2. Fix: Edits to replication specs for cluster

### Test Summary
1. Create a GEOSHARDED cluster with a single replication spec (zone1) with 3 nodes
2. Update the cluster:
  - Change from 3->5 nodes in zone1
  - Add a new replication spec (zone2) with 5 nodes
3. Assert the replication_id is kept on the 1st replication spec, and a new replication spec is added and both replication specs have 5 nodes

### Fix Summary
The fix reads the current cluster if replication specs are set and "matches" replication IDs.
See the unit tests in  [`mappings_test`](https://github.com/mongodb/mongodbatlas-cloudformation-resources/blob/CLOUDP-256534_replication_id_matching/cfn-resources/cluster/cmd/resource/mappings_test.go#L26) for the matching details.

Note: We couldn't reuse the Terraform implementation since CFN doesn't store the full state of the resource.
It only receives the template used for the update call; this means the replication_spec IDs will never be set on the `prevModel` or `currentModel` (this behavior was "discovered" by trying an initial fix of updating the replication specs of the `currentModel` during create)

### Bug reproduced & fix screenshots

Example of [failing e2e test (Github action before fix committed)](https://github.com/mongodb/mongodbatlas-cloudformation-resources/actions/runs/9683006496/job/26717425799)

Screenshot from CFN:

image

Example of [passing e2e test (Github action after fix committed)](https://github.com/mongodb/mongodbatlas-cloudformation-resources/actions/runs/9684226069/job/26721481987)

Screenshot from CFN:

image

Link to any related issue(s): CLOUDP-256730 and CLOUDP-256534

## Type of change:

  expected)

## Manual QA performed:

- [ ] Validated in Atlas UI
- [ ] Included screenshots

## Required Checklist:

  works in Atlas

## Further comments
contract tests:

image

### DIFF
--- a/cfn-resources/cluster/cmd/resource/mappings.go
+++ b/cfn-resources/cluster/cmd/resource/mappings.go
@@ -649,7 +649,7 @@ func setClusterRequest(currentModel *Model) (*admin.AdvancedClusterDescription, 
 
 func PopulateReplicationSpecIds(src, dest []admin.ReplicationSpec) *[]admin.ReplicationSpec {
 	zoneToID := map[string]string{}
-	providerRegionToId := map[string]string{}
+	providerRegionToID := map[string]string{}
 	usedIds := map[string]bool{}
 
 	for _, spec := range src {
@@ -661,7 +661,7 @@ func PopulateReplicationSpecIds(src, dest []admin.ReplicationSpec) *[]admin.Repl
 			zoneToID[zoneName] = specID
 		}
 		if providerRegion := asProviderRegion(spec); providerRegion != "" {
-			providerRegionToId[providerRegion] = spec.GetId()
+			providerRegionToID[providerRegion] = spec.GetId()
 		}
 	}
 	for i, spec := range dest {
@@ -676,7 +676,7 @@ func PopulateReplicationSpecIds(src, dest []admin.ReplicationSpec) *[]admin.Repl
 			dest[i].SetId(idZone)
 			continue
 		}
-		idProvider, foundProvider := providerRegionToId[asProviderRegion(spec)]
+		idProvider, foundProvider := providerRegionToID[asProviderRegion(spec)]
 		providerUsed := usedIds[idProvider]
 		if foundProvider && !providerUsed {
 			usedIds[idProvider] = true
@@ -685,7 +685,6 @@ func PopulateReplicationSpecIds(src, dest []admin.ReplicationSpec) *[]admin.Repl
 		}
 	}
 	return &dest
-
 }
 
 func asProviderRegion(spec admin.ReplicationSpec) string {
--- a/cfn-resources/cluster/cmd/resource/mappings.go
+++ b/cfn-resources/cluster/cmd/resource/mappings.go
@@ -647,7 +647,7 @@ func setClusterRequest(currentModel *Model) (*admin.AdvancedClusterDescription, 
 	return clusterRequest, nil
 }
 
-func PopulateReplicationSpecIds(src, dest []admin.ReplicationSpec) *[]admin.ReplicationSpec {
+func AddReplicationSpecIds(src, dest []admin.ReplicationSpec) *[]admin.ReplicationSpec {
 	zoneToID := map[string]string{}
 	providerRegionToID := map[string]string{}
 	usedIds := map[string]bool{}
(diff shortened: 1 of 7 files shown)


lantoli (approved):

cfn-resources/cluster/cmd/resource/mappings.go
```diff
@@ -646,3 +646,51 @@ func setClusterRequest(currentModel *Model) (*admin.AdvancedClusterDescription,
...
+	zoneToID := map[string]string{}
+	providerRegionToID := map[string]string{}
+	usedIds := map[string]bool{}
+
+	for _, spec := range src {
+		specID := spec.GetId()
+		if specID == "" {
+			continue
+		}
+		if zoneName := spec.GetZoneName(); zoneName != "" {
```
lantoli, line 660: nice, so you're trying to match first by zone name and then by region? does it make sense to fail the update if we're not sure, e.g. we didn't find a matching zone or region?  or that just would mean that it's a new spec and it's ok to go?

EspenAlbert: Exactly, first by zone name and then by region 👍 
Exactly x 2 😄 , if there is no match, that means it is probably a new replication spec

cfn-resources/test/e2e/cluster/cluster_test.go
```diff
@@ -0,0 +1,259 @@
...
+
+func testCreateStack(t *testing.T, c *localTestContext) {
+	t.Helper()
+
+	output := utility.CreateStack(t, c.cfnClient, stackName, c.template)
+	clusterID := getClusterIDFromStack(output)
+
+	project, cluster := readFromAtlas(t, c)
+
+	a := assert.New(t)
```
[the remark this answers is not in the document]
EspenAlbert: 🤔 I like the `assert.New(t)` in this case. I think it works as a "section" header. Although, on delete it makes less sense. Thanks for the comment, I will probably think twice next time 👍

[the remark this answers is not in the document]
lantoli: you can leave it, I think it's just a matter of taste

cfn-resources/cluster/cmd/resource/mappings.go
```diff
@@ -646,3 +646,51 @@ func setClusterRequest(currentModel *Model) (*admin.AdvancedClusterDescription,
...
+	return &dest
+}
+
+func asProviderRegion(spec admin.ReplicationSpec) string {
+	configs := spec.GetRegionConfigs()
+	if len(configs) == 0 {
+		return ""
+	}
+	return fmt.Sprintf("%s-%s", configs[0].GetProviderName(), configs[0].GetRegionName())
+}
```
AgustinBettati, line 696: more as an FYI: there can be replication specs with multiple region configs (case of [multi cloud](https://registry.terraform.io/providers/mongodb/mongodbatlas/latest/docs/resources/advanced_cluster#example-multi-cloud-cluster)). 
My understanding is that if zone name is not defined the cluster is either a replica set or SHARDED with only a single replication spec, so this logic might be redundant.

EspenAlbert: Thanks @AgustinBettati 
Yes, I realized it was a list, but I wasn't sure about the guaranteed order or behavior when adding more region configs.
Therefore, I implemented a "naive" solution only using the 1st region config in the list. 

WDYT? Would it make sense to match on multiple region configs? Is there any guarantee on ordering?

maastha: I don't think API guarantees order so it's better to not rely on that

maastha: I don't think the API guarantees order, it might be better to not rely on that

maastha (approved):

cfn-resources/cluster/cmd/resource/mappings.go
EspenAlbert: An alternative would be to order by the Provider & RegionName and concat these together. 
However, I think keeping it as is works better:

1. I believe there can be a case where the user **adds** an additional regionConfig, if we match on all region configs then the replication ID will never match since the "new" regionConfig will not exist in the currentCluster
2. Keeps the matching logic simpler. 

Happy to hear suggestions: @maastha @AgustinBettati 😊

EspenAlbert: Ok. Choosing to merge as is. Thanks for all the feedback ❤️

PR #1126, opened 2024-08-06
fix: Avoids dumping zero values for Cluster electable specs in json that leads to API validation errors

EspenAlbert:
## Proposed changes

Avoids dumping zero values in json that leads to API validation errors

Link to any related issue(s): CLOUDP-265793

## Type of change:

  expected)

## Manual QA performed:

- [ ] Validated in Atlas UI
- [ ] Included screenshots

## Required Checklist:

- [ ] I have checked that this change does not generate any credentials and that **they are NOT accidentally logged anywhere**.
  works in Atlas


cfn-resources/cluster/cmd/resource/mappings_test.go
```diff
@@ -79,3 +80,39 @@ func regionConfig(provider, region string) *[]admin.CloudRegionConfig {
 		ProviderName: &provider,
 	}}
 }
+
+func TestNewHardwareSpec(t *testing.T) {
+	testCases := map[string]struct {
+		expected string
```
AgustinBettati, line 86: q: is there a reason to not use admin.HardwareSpec for expected value? Just thinking adding the marshaling aspect makes it harder to maintain and not seeing a clear benefit.

cfn-resources/cluster/mongodb-atlas-cluster.json
```diff
@@ -82,7 +82,7 @@
       "type": "object",
       "properties": {
         "DiskIOPS": {
-          "type": "string",
+          "type": "integer",
```
AgustinBettati, line 85: Is this change related or needed? While the change appears correct, this is a breaking change and will require updating CDK as well.

AgustinBettati (commented):
Changes related to bug fix look good, have some doubt on the disk iops change

cfn-resources/cluster/cmd/resource/mappings_test.go
EspenAlbert: More compact and tests the behavior of marshaling that the API call will use. But I'm open to changing it

cfn-resources/cluster/mongodb-atlas-cluster.json
EspenAlbert: This is true. Thank you for calling it out.
I don't see a reason for why it is a string. Both resources in the TF provider use Integer.
Although it is a breaking change, it will only impact users upgrading the CDK/CFN resource, and the fix should be simple.

cfn-resources/cluster/cmd/resource/mappings_test.go
AgustinBettati: > tests the behavior of marshaling that the API call will use

I believe this falls more in the scope of the SDK implementation, we have control over how we populate the SDK model and have understanding that if a nil value is provided for a pointer it is not sent in the request. Don't see this too relevant so adjust as you see fit.

cfn-resources/cluster/mongodb-atlas-cluster.json
Zuhairahmed: suggest leave as is for now, if customer ever asks about it we can revisit. rather then forcing parity between the tools, once something is out customers in that community may already be used to it working this way so would lean on developer/customer empathy here in avoiding breaking changes when possible.

cfn-resources/cluster/cmd/resource/mappings_test.go
EspenAlbert: choosing to keep the code as is. In this case we would be notified if the SDK would change unexpectedly.

cfn-resources/cluster/mongodb-atlas-cluster.json
EspenAlbert: Reverted in be5fdfe361cfd5df527ca9418c4ccb3594c320c3 and re-implemented in conversion in d1c504df5ec4e0ca8dbaf3a4ac01014383114dab

AgustinBettati (approved):
LGTM, thanks for followup changes and good conversations

PR #1661, opened 2026-06-01
chore: Fix service-account contract test List handler violations

lantoli:
## Proposed changes

_Jira ticket:_ CLOUDP-343669

Fix the `service-account` contract test failures by disabling the List handler, following the same pattern as `alert-configuration`.

**Root cause:** The `contract_create_list` and `contract_update_list` tests fail because `SecretExpiresAfterHours` is a `createOnly` property not returned by the Atlas list API. The Create/Update responses preserve it from `currentModel` (value: 720), while List builds each model from an empty struct (value: nil), causing the cfn-cli comparison to fail.

**Why `writeOnlyProperties` does not work:** Adding `SecretExpiresAfterHours` to `writeOnlyProperties` in the schema triggers two contradictory cfn-cli behaviours: (1) `assert_write_only_property_does_not_exist` strictly requires the property to be absent from the response; (2) the request-vs-response comparison tries to prune writeOnly properties by traversing the path `/properties/SecretExpiresAfterHours` as a literal key path in the flat resource model dict, which raises a `LookupError` (`document['properties']` does not exist) and leaves the property in the request, so the comparison still sees 720 (request) vs nil (response) and fails. Both clearing and preserving the value cause failures in different assertions, with no combination that satisfies both.

**Fix:** Disable the List handler (returns `NotFound`, matching `alert-configuration`) and remove `list` from the schema `handlers` section so the list contract tests are skipped. The `handleList` implementation is removed as it is no longer reachable.

Link to any related issue(s): CLOUDP-343669

## Type of change:

  expected)

## Manual QA performed:

- [ ] Validated in Atlas UI
- [ ] Included screenshots

## Required Checklist:

- [x] I have run `make fix` and verified my code
  works in Atlas

## Further comments

### DIFF
--- a/cfn-resources/service-account/cmd/resource/resource.go
+++ b/cfn-resources/service-account/cmd/resource/resource.go
@@ -74,9 +74,9 @@ func Delete(req handler.Request, prevModel *Model, model *Model) (handler.Progre
 }
 
 func List(req handler.Request, prevModel *Model, model *Model) (handler.ProgressEvent, error) {
-	client, setupErr := setupRequest(req, model, ListRequiredFields)
-	if setupErr != nil {
-		return *setupErr, nil
-	}
-	return handleList(client, model), nil
+	return handler.ProgressEvent{
+		OperationStatus:  handler.Failed,
+		Message:          "List operation is not supported",
+		HandlerErrorCode: "NotFound",
+	}, nil
 }
--- a/cfn-resources/service-account/mongodb-atlas-serviceaccount.json
+++ b/cfn-resources/service-account/mongodb-atlas-serviceaccount.json
@@ -137,8 +137,5 @@
     "tagging": {
         "taggable": false
     },
-    "additionalProperties": false,
-    "writeOnlyProperties": [
-        "/properties/SecretExpiresAfterHours"
-    ]
+    "additionalProperties": false
 }
--- a/cfn-resources/service-account/mongodb-atlas-serviceaccount.json
+++ b/cfn-resources/service-account/mongodb-atlas-serviceaccount.json
@@ -127,11 +127,6 @@
             "permissions": [
                 "secretsmanager:GetSecretValue"
             ]
-        },
-        "list": {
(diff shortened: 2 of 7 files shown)


cfn-resources/service-account/cmd/resource/resource.go
```diff
@@ -74,9 +74,9 @@ func Delete(req handler.Request, prevModel *Model, model *Model) (handler.Progre
...
 func List(req handler.Request, prevModel *Model, model *Model) (handler.ProgressEvent, error) {
-	client, setupErr := setupRequest(req, model, ListRequiredFields)
-	if setupErr != nil {
-		return *setupErr, nil
-	}
-	return handleList(client, model), nil
+	return handler.ProgressEvent{
+		OperationStatus:  handler.Failed,
+		Message:          "List operation is not supported",
+		HandlerErrorCode: "NotFound",
```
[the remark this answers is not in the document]
lantoli: follow alert-configuration

cfn-resources/service-account/mongodb-atlas-serviceaccount.json
```diff
@@ -1,141 +1,136 @@
...
-  "description": "Creates and manages a Service Account for an organization.",
-  "sourceUrl": "https://github.com/mongodb/mongodbatlas-cloudformation-resources",
-  "definitions": {
-    "Secret": {
-      "type": "object",
-      "properties": {
-        "Id": {
-          "type": "string",
-          "description": "Unique identifier of the secret."
+    "typeName": "MongoDB::Atlas::ServiceAccount",
```
EspenAlbert, line 2: re-ordering only?

EspenAlbert (approved):
Thanks!

cfn-resources/service-account/mongodb-atlas-serviceaccount.json
lantoli: yes, and removing List handler

